PR #1, opened 2017-03-08
add license

jordimassaguerpla:


cyphar:
How about we add the standard license headers as comments to the `.yaml` files and grains?

mssola:
Also, please add the header in the README.md file like we are doing for the other projects 😉

jordimassaguerpla:
@cyphar : we've never added the license headers to yaml files ... I don't know the reason but I am guessing there must be.

jordimassaguerpla:
@mssola : README contains now the license

mssola:
> @cyphar : we've never added the license headers to yaml files ... I don't know the reason but I am guessing there must be one.

Actually there's none, and we should probably fix this in other repos 😄

jordimassaguerpla:
@mssola : Where can I get an example of the "license headers"?

jordimassaguerpla:
@mssola @cyphar : I added the license headers

jordimassaguerpla:
@mssola @cyphar : does it look good?

ereslibre:
I just wanted to add a comment here, `yaml` files are ok with a license, but maybe `conf` files is too much? Just my 2 cents here.

PR #4, opened 2017-03-09
Add velum configuration settings

ereslibre:
Some settings needs to be mounted from the filesystem, so we can generate this settings for each installation.

Instead of using environment variables, generate `secrets.yml` and `database.yml` files that will be mounted with some generated content for each installation.

This way we don't need to modify manifests (we cannot do it anyway).

There are still some things to adapt.

### DIFF
--- a/salt.yaml
+++ b/salt.yaml
@@ -52,22 +52,22 @@ spec:
   volumes:
     - name: salt-master-config
       hostPath:
-        path: $TODO_PREFIX/config/salt/master.d # FIXME: prefix for this package (config folder)
+        path: /usr/share/caasp-container-manifests/config/salt/master.d
     - name: salt-master
       hostPath:
-        path: $TODO_SALT_STATES # FIXME: where are salt states installed in microos?
+        path: /usr/share/salt/kubernetes/salt
     - name: salt-pillar
       hostPath:
-        path: $TODO_SALT_PILLAR # FIXME: where is salt pillar installed in microos?
+        path: /usr/share/salt/kubernetes/pillar
     - name: salt-sock-dir
       hostPath:
-        path: $TODO_SALT_MASTER_SOCK_DIR # FIXME: where is the salt master sock dir located in microos?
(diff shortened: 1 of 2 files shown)


velum.yaml
```diff
@@ -52,26 +51,26 @@ spec:
...
     env:
     - name: MYSQL_ROOT_PASSWORD # FIXME: do not set root password, create user/password
       value: "salt"             # FIXME
+  volumes:
+    - name: secrets-config
+      hostPath:
+        path: SOMEWHERE/secrets.yml # FIXME: define SOMEWHERE
+    - name: database-config
+      hostPath:
+        path: SOMEWHERE/database.yml # FIXME: define SOMEWHERE
```
jimmykarily: the docker image we use for MariaDB accepts an environment variable (MYSQL_ROOT_PASSWORD) which sets (or resets) the root password to the specified value. We can also use an environment variable in the database.yml file of velum to set the password used. The only question I think is which script will generate the value for those env variables so it is unique among the different installations.

salt.yaml
```diff
@@ -52,22 +52,22 @@ spec:
...
-        path: $TODO_SALT_PILLAR # FIXME: where is salt pillar installed in microos?
+        path: /usr/share/salt/kubernetes/pillar
     - name: salt-sock-dir
       hostPath:
-        path: $TODO_SALT_MASTER_SOCK_DIR # FIXME: where is the salt master sock dir located in microos?
+        path: /var/run/salt/master/sock-dir
     - name: salt-minion-ca-grains
       hostPath:
-        path: $TODO_PREFIX/config/salt/grains/ca # FIXME: prefix for this package (config folder)
+        path: /usr/share/caasp-container-manigests/config/salt/grains/ca
```
ereslibre: Typo here, `manigests` -> `manifests`

ereslibre:
Should this get merged? @jimmykarily @jordimassaguerpla?

jimmykarily:
You got my 👍 and a :beer:

PR #7, opened 2017-03-15
Give mariadb container persistent storage mapped to host /var/lib/mysql

HouzuoGuo:
ReadOnlyOS image 19.2 is the first revision to have the `/var/lib/mysql` subvolume.

### DIFF
--- a/packaging/suse/make_spec.sh
+++ b/packaging/suse/make_spec.sh
@@ -56,7 +56,7 @@ and velum containers on a controller node.
 %build
 
 %install
-for file in salt.yaml velum.yaml; do
+for file in salt.yaml velum.yaml activate.sh; do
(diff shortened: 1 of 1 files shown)


velum.yaml
```diff
@@ -75,3 +75,10 @@ spec:
     env:
     - name: MYSQL_ROOT_PASSWORD # FIXME: do not set root password, create user/password
       value: "salt"             # FIXME
+    volumeMounts:
+      - mountPath: /var/lib/mysql
+        name: velum-mariadb-data
+  volumes:
+    - name: velum-mariadb-data
+      hostPath:
```
mssola: I thought that these manifests were meant for production-ready deployments. If so refrain from using `hostPath`. That being said, maybe you could just add a comment here and add a note in the README or something ?

ereslibre: We don't have any other place to persist ATM :(

PR #10, opened 2017-03-15
[WIP] add pv-recycler-node image

jordimassaguerpla:


jordimassaguerpla:
@ereslibre : what about it now?

ereslibre:
LGTM, maybe we should wait for others to dump their thoughts on this :smile:

jordimassaguerpla:
Should we add the option here? https://github.com/kubic-project/salt/blob/master/salt/kubernetes-master/config.jinja and https://github.com/kubic-project/salt/blob/master/salt/kubernetes-minion/config.jinja

ereslibre:
From what I can see, we do:

```
EnvironmentFile=/etc/kubernetes/config (ignore_errors=yes)
EnvironmentFile=/etc/kubernetes/controller-manager (ignore_errors=yes)
```

at the `kube-controller-manager.service`. Now that I see it actually [the comment here](https://github.com/kubic-project/salt/blob/master/salt/kubernetes-minion/config.jinja) isn't accurate, but yes, we need to adapt the flags [here](https://github.com/kubic-project/salt/blob/master/salt/kubernetes-master/config.jinja) to refer to the pv-recycler template that we are adding.

mssola:
Note that this files *must* be present before starting the controller manager, and adding this manifest file in this repo assumes that Kubernetes is already running, which is contradictory. Therefore, this manifest file should go into the salt repo instead, where we'll provision the kubernetes master with this manifest, and add the flag to the controller manager as we used to do when we had SLE12 support.

ereslibre:
Actually this package will come preinstalled with MicroOS, so this is a step before salt runs, but it makes sense to add it on salt level instead of here (I don't have a strong opinion on where to provide it).

ereslibre:
Scratch that, you are absolutely right @mssola. This needs to be done at salt level.

ereslibre:
It's better at salt level to make this work on opensuse deployments too, but for MicroOS it won't make a difference (the yaml will come preinstalled). +1 for moving it to our salt states.

jordimassaguerpla:
https://github.com/kubic-project/salt/pull/19

PR #11, opened 2017-03-17
Add activate

jordimassaguerpla:


activate.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+cd $(dirname $0)
+kube_dir=/etc/kubernetes/manifests
+if [ ! -d $kube_dir ];then
```
ereslibre: Style, space after `;`

ereslibre:
My only concern is if we can `cp` at this point or is this still RO and we would need to request this folder to be RW.

jordimassaguerpla:
I just added some comments received by email.

PR #14, opened 2017-03-20
add the required images for caasp as "Requires"

jordimassaguerpla:


jordimassaguerpla:
Hi. This is very straightforward and I'd like to submit this package so it can be reviewed ASAP so that we have a new iso 2morrow morning.

PR #18, opened 2017-03-22
salt-minion is not activated in the controller node

jordimassaguerpla:
thus there is no need to disabled it


jordimassaguerpla:
Included in https://github.com/kubic-project/caasp-container-manifests/pull/19

PR #23, opened 2017-03-23
Use port 80 by default

ereslibre:
Use port 80 by default for velum service until we can actually use
port 443 to serve this content.


jordimassaguerpla:
lgtm

PR #24, opened 2017-03-23
review entry commands

jordimassaguerpla:


velum.yaml
```diff
@@ -69,7 +69,7 @@ spec:
     - name: VELUM_SALT_USER
       value:
     command: ["/bin/sh","-c"]
-    args: ["bundle exec bin/rake salt:process"]
+    args: ["$BUNDLER_BIN exec bin/rake salt:process"]
```
ereslibre: This one should be kept as it was.

jordimassaguerpla: bundle does not have to be in the PATH in the image. If bundler is vendored (as in /srv/velum/vendor), we need BUNDLER_BIN to be setup.

PR #28, opened 2017-03-27
Added a temporary fix for the pause container in the dashboard

mssola:
This fix should eventually go away once we refactor how we provision the
dashboard node.

Signed-off-by: Miquel Sabaté Solà <msabate@suse.com>

### DIFF
--- a/activate.sh
+++ b/activate.sh
@@ -22,6 +22,10 @@ fi
 cp -v $manifest_dir/salt.yaml $kube_dir
 cp -v $manifest_dir/velum.yaml $kube_dir
 
+# Make sure that the controller node looks for the local pause image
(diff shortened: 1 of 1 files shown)


activate.sh
```diff
@@ -22,6 +22,9 @@ fi
 cp -v $manifest_dir/salt.yaml $kube_dir
 cp -v $manifest_dir/velum.yaml $kube_dir
 
+# Make sure that the controller node looks for the local pause image
```
inercia: A `TODO` would remind us to remove this aberration... :-)

mssola: Done

jordimassaguerpla: +1

jordimassaguerpla: lgtm because it is the minimal fix for this ...

PR #29, opened 2017-03-28
Skip copies when the manifests already exist there

inercia:
We already have _development versions_ of the manifest files at `/etc/kubernetes/manifests` when using Terraforrm, so we should skip the copies performed by `activate.sh`. Once we have this merged, we will be able to run `activate.sh` from `provision-dashboard.sh`

### DIFF
--- a/activate.sh
+++ b/activate.sh
@@ -1,26 +1,40 @@
 #!/bin/sh
+
 kube_dir=/etc/kubernetes/manifests
 manifest_dir=/usr/share/caasp-container-manifests
-if [ ! -d $kube_dir ]; then
-    echo "$kube_dir does not exist" >&2
-    echo "make sure kubernetes is installed" >&2
-    exit -1
-fi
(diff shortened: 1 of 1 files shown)


ereslibre:
Wouldn't this break manifest updates?

inercia:
> Wouldn't this break manifest updates?

You are right: I skip the copies only when some flag is provided...

jordimassaguerpla:
@inercia : Could we have some default values for the dirs? YaST is calling this script without any argument at all.

inercia:
> Could we have some default values for the dirs? YaST is calling this script without any argument at all.

@jordimassaguerpla The default values for dirs are still there :-)

jordimassaguerpla:
@inercia I overlooked that.
Please wait for merging this thought.

inercia:
Closing in favour of some equivalent code in [`provision/dashboard.sh`](https://github.com/kubic-project/terraform/blob/master/provision/dashboard.sh)...

PR #32, opened 2017-04-10
activate.sh: fix bsc#1032651

MaximilianMeister:
docker, kubelet and etcd on the administration node are now enabled by
the system role, we don't need to do that anymore in the activate.sh
script.

https://bugzilla.suse.com/show_bug.cgi?id=1032651


MaximilianMeister:
> Also, depending on packaging, we can also remove this.

not sure about that, @jordimassaguerpla do you know if we can drop this as well?

jordimassaguerpla:
About the change @ereslibre pointed out, we can not change that in the package because the package needs to remain generic (i.e. this is the same package for both opensuse and sles)

PR #34, opened 2017-04-24
Migrate manifests to `public/private` scheme

ereslibre:


ereslibre:
Can you check now?

packaging/suse/make_spec.sh
```diff
@@ -64,13 +64,13 @@ and velum containers on a controller node.
...
 %install
-for file in salt.yaml velum.yaml; do
+for file in public.yaml private.yaml; do
   install -D -m 0644 \$file %{buildroot}/%{_datadir}/%{name}/\$file
 done
 install -D -m 0755 activate.sh %{buildroot}/%{_datadir}/%{name}/activate.sh
-for dir in grains master.d minion.d-ca; do
-  install -d %{buildroot}/%{_datadir}/%{name}/config/salt/\$dir
-  install config/salt/\$dir/* %{buildroot}/%{_datadir}/%{name}/config/salt/\$dir
+for dir in mysql salt/grains salt/master.d salt/minion.d-ca; do
```
MaximilianMeister: can't say if that works fine, but we'll see in the packaging i guess

ereslibre: Thank you for the hint! :clap:

private.yaml
```diff
@@ -0,0 +1,41 @@
...
+  labels:
+    name: velum-private
+spec:
+  hostNetwork: false
+  containers:
+  - name: velum-mariadb
+    image: sles12/mariadb:10.0
+    env:
+    - name: MYSQL_ROOT_PASSWORD
+      value: "salt"
```
robdaemon: This still makes me uncomfortable in a production environment. We should ideally generate a random password.

public.yaml
```diff
@@ -53,7 +58,66 @@ spec:
...
+    - name: VELUM_DB_SOCKET
+      value: /var/run/mysqld/mysqld.sock
+    - name: VELUM_SALT_HOST
+      value: "127.0.0.1"
+    - name: VELUM_SALT_PORT
+      value: "8000"
+    - name: VELUM_SALT_USER
+      value: saltapi
+    - name: VELUM_SALT_PASSWORD
+      value: saltapi
```
robdaemon: Same with this password :(

private.yaml
ereslibre: I completely agree, but IMHO we should do this in a dedicated infra secrets PR, where we generate random infra secrets (mysql root password, salt api user & password) and persist them for that installation.

public.yaml
ereslibre: Same, we should do this in a dedicated PR IMHO...

private.yaml
robdaemon: :+1: do we already have a card or bug entered for this?

ereslibre: I created https://trello.com/c/VB40KK1s/211-generate-infrastructure-secrets, feel free to expand or improve it as you see fits :)

ereslibre: @robdaemon I created it as we spoke here, that's why you couldn't find it :)

PR #38, opened 2017-05-03
Replace the mount points for Salt code and pillar

inercia:
Change the mount points for Salt code and pillars, so we can either install the Salt code from RPMs **or** mount those paths from a Salt repo checkout.

### DIFF
--- a/config/salt/master.d/salt-master.conf
+++ b/config/salt/master.d/salt-master.conf
@@ -1,8 +1,8 @@
 user: root
 auto_accept: True
 interface: 0.0.0.0
-
 event_return: mysql
+presence_events: True
 
 file_roots:
   base:
--- a/config/salt/master.d/salt-master.conf
+++ b/config/salt/master.d/salt-master.conf
@@ -4,6 +4,14 @@ interface: 0.0.0.0
 
 event_return: mysql
 
+file_roots:
+  base:
+    - /usr/share/salt/kubernetes/salt
+
+pillar_roots:
+  base:
+    - /usr/share/salt/kubernetes/pillar
+
 mysql:
   # salt does not support specifying the UNIX socket location here - as a workaround,
   # use the MYSQL_UNIX_PORT environment variable used by libmysqlclient
--- a/public.yaml
+++ b/public.yaml
@@ -30,10 +30,8 @@ spec:
       name: salt-master-config
     - mountPath: /etc/salt/pki/master
       name: salt-master-pki
-    - mountPath: /srv/salt
-      name: salt-master
-    - mountPath: /srv/pillar
-      name: salt-pillar
+    - mountPath: /usr/share/salt/kubernetes
+      name: salt
     - mountPath: /var/run/salt/master
       name: salt-sock-dir
     - mountPath: /var/run/mysql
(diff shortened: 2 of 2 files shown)


config/salt/master.d/salt-master.conf
```diff
@@ -1,8 +1,16 @@
 user: root
 auto_accept: True
 interface: 0.0.0.0
-
 event_return: mysql
+presence_events: True
```
ereslibre: I am a bit worried this can spam our `salt_events` table...

inercia: We will need this anyway for detecting nodes that have gone down...

inercia:
> After giving some thought to this change... why would we need to do this if the salt-master is always going to run as a container?

@ereslibre This would be more a transitional change until we properly install the Salt code in the container. Until then, nothing will break.

> We could just mount /srv/pillar and /srv/salt since they are the default locations and we wouldn't need to specify their locations on the salt-master.conf, no?

We specify `/usr/share/salt/kubernetes` because that is the installation root where everything from  `kubernetes-salt.rpm` is installed. This way we can use both approaches: rpm in the image or mounts. Then we could drop one of them if we want...

ereslibre:
From a production point of view the current code is just fine, in the sense that we have a salt-master container, where we mount `/srv/salt` from `/usr/share/salt/kubernetes/salt` and `/srv/pillar` from `/usr/share/salt/kubernetes/pillar`. I don't see a reason to change that on the salt-master container... that also forces us to modify the default configuration salt and pillar locations.

Why is this change required? I might be lacking some context here...

inercia:
> From a production point of view the current code is just fine, in the sense that we have a salt-master container, where we mount /srv/salt from /usr/share/salt/kubernetes/salt and /srv/pillar from /usr/share/salt/kubernetes/pillar. I don't see a reason to change that on the salt-master container... that also forces us to modify the default configuration salt and pillar locations.

@ereslibre Because we will not mount anything, and the RPM will not install the {salt,pillar} in `/srv/{salt,pillar}` but on `/usr/share/salt/kubernetes/{salt,pillar}`. So once we create a `salt-master` image with something like:

```
FROM sle...
RUN zypper in salt-master kubernetes-salt
```

we have to point the paths. And changing the states and pillar locations is trivial...

ereslibre:
Another possibility is to remove kubernetes-salt from the administration dashboard and the installation in general (so salt-states and pillar are only inside the salt-master) and this package installs the salt states and pillar on the default location, so there's no need to change the salt-master configuration.

ereslibre:
However I am not fully convinced at the moment, since `kubernetes-salt` already comes preinstalled... so why install it twice if we can just mount it from the host where it comes preinstalled (and we will need to keep it preinstalled because of the workers -- and we only want one unique set of packages system wide).

inercia:
> However I am not fully convinced at the moment, since kubernetes-salt already comes preinstalled... so why install it twice if we can just mount it from the host where it comes preinstalled (and we will need to keep it preinstalled because of the workers -- and we only want one unique set of packages system wide).

The answer is very simple: it will not come preinstalled.

ereslibre:
> The answer is very simple: it will not come preinstalled. :-)

:+1:

PR #39, opened 2017-05-03
activate.sh: notify that velum is starting (bsc#1031682)

MaximilianMeister:
we choose to do that in activate.sh instead of baking it into the
MicroOS, because it only refers to the controller/admin node

https://bugzilla.suse.com/show_bug.cgi?id=1031682

welcome

### DIFF
--- a/activate.sh
+++ b/activate.sh
@@ -29,3 +29,10 @@ sed -i 's|--config=/etc/kubernetes/manifests|--config=/etc/kubernetes/manifests 
 # Make sure etcd listens on 0.0.0.0
 sed -i 's@#\?ETCD_LISTEN_PEER_URLS.*@ETCD_LISTEN_PEER_URLS=http://0.0.0.0:2380@' /etc/sysconfig/etcd
 sed -i 's@#\?ETCD_LISTEN_CLIENT_URLS.*@ETCD_LISTEN_CLIENT_URLS=http://0.0.0.0:2379@' /etc/sysconfig/etcd
+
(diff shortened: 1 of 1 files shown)


MaximilianMeister:
> Shouldn't we say something about the possible delay until the Dasboard is up and running?

yes, but that will be in the installer, as it only affects the installation stage. the yast team will take care about that

MaximilianMeister:
@ereslibre btw using `issue-generator` is only meant for ssh keys and network addresses, so we couldn't use it here as it is a generic message

jordimassaguerpla:
@MaximilianMeister : what is "issue-generator"?
@ereslibre : We had tried other times to print the IP address in the issue but that proved to be a bad idea because if you are using dhcpd, you may not yet have the ip address when the issue gets printed. You could instead create a service that waits for the ip address to be set and then prints the message, unless there is a way for the issue-generator to wait on the ip address.

ereslibre:
`/etc/issue` usually supports what `getty` supports (`man issue`). For example, in my system:

```
Welcome to openSUSE Tumbleweed 20170424 - Kernel \r (\l).


enp0s31f6: \4{enp0s31f6} \6{enp0s31f6}
```

ereslibre:
We can improve this in the future though...

MaximilianMeister:
yeah, but getting the ip address seems to need a helper script like `issue-generator`, we'd have to adapt that one i guess

ereslibre:
Let's do it in the near future if we see if needed :)

PR #40, opened 2017-05-05
Share ssh public key for autoyast profile, bsc#1030876

mjura:


activate.sh
```diff
@@ -36,3 +36,11 @@ You can manage your cluster by opening the web application running on
...
 
 EOF
+
+# Generate root ssh key and share it with velum
+# https://bugzilla.suse.com/show_bug.cgi?id=1030876
+if ! [ -f /root/.ssh/id_rsa ]; then
+  ssh-keygen -b 2048 -f /root/.ssh/id_rsa -t rsa -N ""
+fi
+[ -d /tmp/ssh ] || mkdir /tmp/ssh
+cp /root/.ssh/id_rsa.pub /tmp/ssh
```
ereslibre: Our intention is to kill the `activate.sh` scrfipt. We should move the key generation somewhere else, no?

mjura: This was discussed on caasp mailing list, and I tried also submit this change ttps://build.suse.de/request/show/132262. Thorsten said that this should be done in`activate.sh` script

mjura:
> Why /tmp? I think we should generate the key somewhere else if we don't want to lose it after a reboot or a /tmp cleanup. We would lose access to the machines we were granted access with autoyast if we reboot the admin node...

Ok, I agree. What is your proposal then?

ereslibre:
@mjura I haven't looked very closely at what is the best RW place... but maybe we can use `/var/lib` for this? Or `/etc`?

activate.sh
ereslibre: Sorry, missed that. Let's add it here then for now, but I fear we will have to wipe this from here at some point...

mjura:
@ereslibre: ok, I've change path for this ssh public key on `/var/lib/misc/ssh-public-key`. Please let me know, if this is ok for you. Thank you

activate.sh
```diff
@@ -36,3 +36,11 @@ You can manage your cluster by opening the web application running on
 port 80 of this node from your browser.
 
 EOF
+
+# Generate root ssh key and share it with velum
+# https://bugzilla.suse.com/show_bug.cgi?id=1030876
+if ! [ -f /root/.ssh/id_rsa ]; then
+  ssh-keygen -b 2048 -f /root/.ssh/id_rsa -t rsa -N ""
```
ereslibre: We could generate a 4096 bits key by default.

mjura: Done, I change this to 4096 bits key

PR #42, opened 2017-05-09
[WIP] add init containers

PI-Victor:
Based on the comment from https://github.com/kubic-project/velum/pull/151 i've moved the init containers here. it's working quite ok except for a few snags which i'm trying to resolve.

### DIFF
--- a/public.yaml
+++ b/public.yaml
@@ -19,12 +19,61 @@ metadata:
     name: velum-public
 spec:
   hostNetwork: true
+  initContainers:
+  - name: init-mariadb
+    image: library/mariadb:10.0.29
+    imagePullPolicy: IfNotPresent
+    env:
+    - name: MYSQL_ROOT_PASSWORD
+      value: "salt"
+    - name: MARIADB_SOCKET
+      value: /var/run/mysqld/mysqld.sock
+    command: ["/bin/sh", "-c"]
+    args: ["until mysqladmin -p\"${MYSQL_ROOT_PASSWORD}\" -S \"${MARIADB_SOCKET}\" ping; do echo 'Database not online yet, waiting...' && sleep 3; done"]
+    volumeMounts:
+      - mountPath: /var/run/mysqld
+        name: mariadb-unix-socket
+  - name: init-rake-setup
+    image: opensuse/velum:development
+    imagePullPolicy: IfNotPresent
+    env:
+    - name: RAILS_ENV
+      value: "development"
+    - name: VELUM_SECRETS_DIR
+      value: /var/lib/misc/velum-secrets
+    - name: RAILS_SERVE_STATIC_FILES
+      value: "true"
+    - name: VELUM_PORT
+      value: "3000"
+    - name: VELUM_DB_USERNAME
+      value: "root"
+    - name: VELUM_DB_PASSWORD
+      value: "salt"
+    - name: VELUM_DB_SOCKET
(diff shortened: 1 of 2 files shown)


PI-Victor:
@inercia yes, i will re-add the readiness probes i wanted to add back in https://github.com/kubic-project/velum/pull/137/files
it would be nice if velum had a /_healtz endpoint that returns 200 when the app is up.
i actually wanted to add that, but i'm not a ruby dev and have no intention to learn ruby at any given time :D. Here's what i'm talking about https://github.com/kelseyhightower/app-healthz
cc: @ereslibre for healtz endpoint :)

mssola:
> it would be nice if velum had a /_healtz endpoint that returns 200 when the app is up.
i actually wanted to add that, but i'm not a ruby dev and have no intention to learn ruby at any given time :D. Here's what i'm talking about https://github.com/kelseyhightower/app-healthz

Easy as pie: https://github.com/kubic-project/velum/pull/156 😉

public.yaml
```diff
@@ -19,12 +19,61 @@ metadata:
...
+    - name: VELUM_SALT_HOST
+      value: "127.0.0.1"
+    - name: VELUM_SALT_PORT
+      value: "8000"
+    - name: VELUM_SALT_USER
+      value: "saltapi"
+    - name: VELUM_SALT_PASSWORD
+      value: "saltapi"
+    command: ["/bin/sh", "-c"]
+    args: ["bundle exec rake db:create && bundle exec rake db:schema:load && bundle exec rake db:migrate"]
```
ereslibre: This will break the development environment, maybe we need to split `bin/init` in https://github.com/kubic-project/velum/blob/master/bin/init, but the `db:schema:load` needs this https://github.com/kubic-project/velum/blob/feac9eb14c16d0e250feeeb2e42c73a64781e178/bin/init#L41-L43.

If we split the start script we could call that script from here, without the need to explicitly run this commands, and it would work fine in development and in production.

public.yaml
```diff
@@ -72,47 +121,46 @@ spec:
...
-      value: saltapi
+      value: "saltapi"
     volumeMounts:
       - mountPath: /var/run/mysql
         name: mariadb-unix-socket
       - mountPath: /var/lib/misc/velum-secrets
         name: velum-secrets
-    command: ["/bin/sh","-c"]
-    args: ["bin/init"]
+    command: ["/bin/sh", "-c", "bundle exec rails s -b 0.0.0.0 -p $VELUM_PORT --pid /tmp/puma.pid Puma"]
```
ereslibre: Improving `bin/init` would help here too... we could do something like `bin/init --skip-initialization` or something like that... that way we avoid duplicating commands like this one...

PI-Victor:
kubelet manifests only support pod specs and don't support configmaps and other things.
it would be nice if we could deploy velum, not as a pod spec, but as a full deployment. this way we can take advantage of configmaps and everything else. it would also make updating `velum` really easy
do we have such a plan, am i missing something?

ereslibre:
Yeah... a standalone kubelet (disconnected -- no apiserver) as it is running in the administration dashboard supports a minimal subset of features...

mssola:
@PI-Victor just merged the PR providing the `/_health` endpoint. You are free to use it now 😉

PI-Victor:
@mssola  thanks, i will create a live/readiness probe with that endpoint!

PI-Victor:
closing this for now.

PR #43, opened 2017-05-16
Mount MariaDB configuration under `/etc/my.cnf.d`

ereslibre:
* Under SLE the configuration lives under `/etc/my.cnf.d`
* Add `[mysqld]` section to the skip-networking file so it will be
  processed by mysqld (otherwise it's ignored)
* Mount only the `skip-networking.cnf` file, as other cnf files come
  pre-installed in `/etc/my.cnf.d` and we would be shadowing them

### DIFF
--- a/config/mysql/skip-networking.cnf
+++ b/config/mysql/skip-networking.cnf
@@ -1,1 +1,2 @@
+[mysqld]
 skip-networking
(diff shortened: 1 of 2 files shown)


ereslibre:
I think I closed this by mistake

PR #44, opened 2017-05-18
Update message to reflect what has been proposed on the bug report

ereslibre:
Fixes: bsc#1031682


ereslibre:
We won't be updating the message yet.

PR #46, opened 2017-05-22
Remove hardcoded secrets

ereslibre:
We will be generating secrets with init containers. These secrets will
be created in a volume mounted from the host, so they survive reboots.

While being sufficient for our GA purposes we will need to rethink how
we do this in a HA environment.

Some secrets are generated with the init containers:

* mysql root password
* mysql velum user password
* mysql salt user password
* saltapi user password

Once we have generated all the passwords, we need to write this
configuration on files that will be mounted on the different containers,
so the different services can read the files where the passwords are written.
By default, passwords will be created in files with permissions 400.

Password generation uses `/dev/random`, performing a `base64` encoding to
that random content, and pick up a line of the `base64` output.

Images will take this environment variables and they will use their
entrypoint to perform the required actions. Example:

* mariadb container will set the root password and do some initializations
* salt-master container will `chpasswd` the `saltapi` user to the generated
  saltapi password.

### DIFF
--- a/packaging/suse/make_spec.sh
+++ b/packaging/suse/make_spec.sh
@@ -75,6 +75,7 @@ for dir in mysql salt/grains salt/minion.d-ca; do
   install -d %{buildroot}/%{_datadir}/%{name}/config/\$dir
   install config/\$dir/* %{buildroot}/%{_datadir}/%{name}/config/\$dir
 done
+install -d %{buildroot}/%{_datadir}/%{name}/setup
 
 
 %files
--- a/private.yaml
+++ b/private.yaml
@@ -17,16 +17,34 @@ metadata:
   name: velum-private
   labels:
     name: velum-private
+  annotations:
+    pod.beta.kubernetes.io/init-containers: '[
+      {
+        "name": "mariadb-secrets",
+        "image": "sles12/velum:0.0",
+        "command": ["sh", "-c", "umask 377;
+                                 if [ ! -f /infra-secrets/mariadb-root-password ]; then
+                                   head -n 10 /dev/random | base64 | head -n 10 | tail -n 1 > /infra-secrets/mariadb-root-password;
+                                 fi;
+                                 exit 0"],
+        "volumeMounts": [
+          {
+            "mountPath": "/infra-secrets",
+            "name": "infra-secrets"
+          }
+        ]
+      }
+    ]'
 spec:
   hostNetwork: false
   containers:
   - name: velum-mariadb
     image: sles12/mariadb:10.0
     env:
-    - name: MYSQL_ROOT_PASSWORD
-      value: "salt"
     - name: MYSQL_DISABLE_REMOTE_ROOT
-      value: "false"
+      value: "true"
+    - name: MYSQL_ROOT_PASSWORD_FILE
+      value: /var/lib/misc/infra-secrets/mariadb-root-password
     volumeMounts:
       - mountPath: /var/lib/mysql
         name: mariadb-data
@@ -34,6 +52,9 @@ spec:
         name: mariadb-unix-socket
       - mountPath: /etc/my.cnf.d/skip-networking.cnf
         name: mariadb-config-skip-networking
+        readOnly: True
+      - mountPath: /var/lib/misc/infra-secrets
+        name: infra-secrets
         readOnly: True
   volumes:
     - name: mariadb-data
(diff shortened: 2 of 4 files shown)


private.yaml
```diff
@@ -17,23 +17,39 @@ metadata:
...
+        "image": "sles12/velum:development",
+        "command": ["sh", "-c", "if [ ! -f /infra-secrets/mariadb-password ]; then head -n 1000 /dev/random | sha256sum | cut -d\" \" -f1 > /infra-secrets/mariadb-password; fi; exit 0;"],
+        "volumeMounts": [
+          {
+            "mountPath": "/infra-secrets",
+            "name": "infra-secrets"
+          }
+        ]
+      }
+    ]'
```
PI-Victor: i think you can use the spec.initContainers unless we're worried that we'll have to rollback to 1.6 for the api server
https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#support-and-compatibility
like i did in the previous PR https://github.com/kubic-project/caasp-container-manifests/pull/42/files

also, i was busy with bugsquad stuff so i didn't have time to finish the previous PR, if you want i can close that PR and you can implement it, since you're more familiar with the setup?
not trying to dump more work on you tho 😆

ereslibre: As far as I could see here https://kubernetes.io/docs/concepts/workloads/pods/init-containers/ this syntax is 1.6 only... for 1.5 (that is what we have in CaaSP right now) only the annotations syntax will work.

When we are stable with 1.6 we should move to the new syntax  though... that is much more convenient...

PI-Victor: ah, we didn't upgrade to `1.6` - i forgot, my bad.

flavio:
This looks like a significant improvement over the current implementation.

@ukbelch what do you think about that from a security POV?

public.yaml
```diff
@@ -17,6 +17,46 @@ metadata:
...
+            "mountPath": "/infra-secrets",
+            "name": "infra-secrets"
+          }
+        ]
+      },
+      {
+        "name": "salt-master-config",
+        "image": "sles12/velum:0.0",
+        "command": ["sh", "-c", "rmdir /salt-master-credentials/returner-credentials.conf; echo \"mysql:\" > /salt-master-credentials/returner-credentials.conf; echo -n \"  pass: \" >> /salt-master-credentials/returner-credentials.conf; cat /infra-secrets/mariadb-password >> /salt-master-credentials/returner-credentials.conf"],
+        "volumeMounts": [
```
PI-Victor: keep wondering if this could be simplified a bit, but to be honest it looks ok for now at least.

ereslibre: Yeah... I am taking a look on how to make this a bit more readable... I agree it's not the most intuitive thing.

PI-Victor:
is this blocked, can it be merged?

ereslibre:
> is this blocked, can it be merged?

I think this is halted until we release the Beta 4, so it's not another point of friction.

ereslibre:
Please, do not merge yet. Commit message needs tweaking, as well as PR description.

ereslibre:
@flavio Asking for your review. Had to add `setup` folder because here documents didn't work fine on the annotations, and the double nesting quoting was making things nearly unreadable.

Please, confirm if everything looks fine for you. Merging this will require merging other related PR's and SR's, namely:

- https://github.com/kubic-project/caasp-devenv/pull/1
- https://github.com/kubic-project/salt/pull/83
- https://build.suse.de/request/show/133060
- https://build.suse.de/request/show/133059
- https://build.suse.de/request/show/133030

ereslibre:
@jordimassaguerpla asking for your review because I'm not sure I got the packaging side of things right.

ereslibre:
> This PR has been created sometimes ago, is everything working fine against rebased master?

Yes, I have been constantly rebasing them to master (all PR's), so everything should be fine.

flavio:
Ok, then let's merge it. Please @ereslibre coordinate with the others to avoid issues during the merge process (eg: breaking the daily image or such).

packaging/suse/make_spec.sh
```diff
@@ -75,6 +75,7 @@ for dir in mysql salt/grains salt/minion.d-ca; do
   install -d %{buildroot}/%{_datadir}/%{name}/config/\$dir
   install config/\$dir/* %{buildroot}/%{_datadir}/%{name}/config/\$dir
 done
+install -d %{buildroot}/%{_datadir}/%{name}/setup
```
jordimassaguerpla: lgtm

PR #48, opened 2017-05-29
review spec file

jordimassaguerpla:
fix some things in the specfile which otherwise they need to be fixed
afterwards


jordimassaguerpla:
This is not a Priority 1 issue, so we can wait to merge this until after beta4 has been released.

mssola:
> This is not a Priority 1 issue, so we can wait to merge this until after beta4 has been released.

I think it can be merged safely. It's not introducing any breaking changes or anything weird into it 👍

PR #49, opened 2017-06-02
Add `readOnly` to mounts from where we will only read

ereslibre:
When the kubelet tries to mount certain mountpoints as RW, it will raise
an error if the underlying host has this directories as RO. We will only
mount as RW the folders where we really need to write, skipping those
warnings from the kubelet.

Related: bsc#1042382

### DIFF
--- a/private.yaml
+++ b/private.yaml
@@ -34,6 +34,7 @@ spec:
         name: mariadb-unix-socket
       - mountPath: /etc/my.cnf.d/skip-networking.cnf
         name: mariadb-config-skip-networking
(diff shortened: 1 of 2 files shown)


flavio:
Please fix this PR to have a meaningful description. Also fix the commits to include references to [bsc#1042382](https://bugzilla.suse.com/show_bug.cgi?id=1042382); this is **really important**.

flavio:
LGTM aside from the issues I pointed above

ereslibre:
@flavio Updated, thank you.

PR #53, opened 2017-06-06
Create MariaDB system user for /var/lib/mariadb, bsc#1034885

mjura:
Create and reserve UID/GID for user MariaDB to own /var/lib/mariadb
for security reasons

### DIFF
--- a/activate.sh
+++ b/activate.sh
@@ -47,6 +47,10 @@ fi
 [ -d /var/lib/misc/ssh-public-key  ] || mkdir -p /var/lib/misc/ssh-public-key
 cp /root/.ssh/id_rsa.pub /var/lib/misc/ssh-public-key
 
+# Create MariaDB system user, bsc#1034885
+groupadd -g 331 mysql
+useradd -r -u 60 -g mysql -c "MySQL database admin" -d /var/lib/mysql -s /bin/false mysql
+
 # Connect the salt-minion running in the administration controller node to the local salt-master
 # instance that is running in a container
 cat <<EOF > /etc/salt/grains


flavio:
To be put on hold until we agree on how to address bsc#1034885

activate.sh
```diff
@@ -46,3 +46,7 @@ if ! [ -f /root/.ssh/id_rsa ]; then
 fi
 [ -d /var/lib/misc/ssh-public-key  ] || mkdir -p /var/lib/misc/ssh-public-key
 cp /root/.ssh/id_rsa.pub /var/lib/misc/ssh-public-key
+
+# Create MariaDB system user, bsc#1034885
+groupadd -g 331 mysql
+useradd -r -u 331 -g mysql -c "MySQL database admin" -d /var/lib/mysql -s /bin/false mysql
```
flavio: Where does these numbers come from? Are these the ones used by default by the mariadb package on openSUSE/SLE?

PI-Victor:
any updates on the workflow, bugzilla discussion didn't progress passed this PR.

activate.sh
jordimassaguerpla: You better use uid 60 for mysql user, which is the one used by mariadb user:

See: https://build.suse.de/package/view_file/SUSE:SLE-12-SP2:Update/mariadb/mariadb.spec?expand=1

You can see in the pre section this:

getent group mysql >/dev/null || groupadd -r mysql
getent passwd mysql >/dev/null || useradd -r -o -g mysql -u 60 -c "MySQL database admin" \
                  -s /bin/false -d %{_localstatedir}/lib/mysql mysql
usermod -g mysql -s /bin/false mysql

jordimassaguerpla: Group does not have a gid in the package so you cannot know in advance which will be choosen in the mariadb image, but you can be certain about the uid 60 .

mjura: @jordimassaguerpla : Thank you for your findings about mariadb. For GID we can modify kiwi mariadb container template and add mysql group in advance with fixed GID.

mjura:
@PI-Victor : I've fixed comments and rebased this change

PI-Victor:
anyone know - can we proceed with merging this? or are we still waiting for the other bug to be solved?

flavio:
We have put this PR on hold, we need to agree about how to fix that on bugzilla.

ereslibre:

PR #55, opened 2017-06-08
Do not mount these three mountpoints readonly

ereslibre:
Related to infrastructure secrets.

It makes the container initialization to fail. Ideally they should be
read-only, as they will only read from here, but something is trying
to write in there, avoiding containers to start.


jordimassaguerpla:
lgtm

PR #56, opened 2017-06-08
Do not mount `/usr/share/salt/kubernetes/config/master.d` from the host

ereslibre:
We will mounting other volumes on top of this on the containers, and
they will fail because on the host, `/usr/share/salt/kubernetes/config/master.d`
is a `RO` volume.

We fix this by mounting all specific files in the containers instead of
the top level directory of the hierarchy. This imposes us the restriction
to modify the container manifests every time a new config file appears, but
that should not happen very often.

Otherwise, we cannot add our own configuration files on top of the `RO`
mounted volume, because they will fail. In this case, the mounted folder
on the containers will be `/etc/salt/master.d`, but in this case this folder
won't be mounted from `/usr/share/salt/kubernetes/config/master.d`, it will
live only in the container, and we will mount the specific files under it,
what will avoid the `RO` volume problems from the host.

### DIFF
--- a/public.yaml
+++ b/public.yaml
@@ -95,10 +95,24 @@ spec:
     - name: SALTAPI_PASSWORD_FILE
       value: /var/lib/misc/infra-secrets/saltapi-password
     volumeMounts:
-    - mountPath: /etc/salt/master.d
-      name: salt-master-config
+    - mountPath: /etc/salt/master.d/master.conf
+      name: salt-master-config-master-conf
+      readOnly: True
+    - mountPath: /etc/salt/master.d/api.conf
+      name: salt-master-config-api-conf
+      readOnly: True
+    - mountPath: /etc/salt/master.d/peer.conf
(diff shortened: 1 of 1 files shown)


jordimassaguerpla:
lgtm

kiall:
From memory, this means the content of the files mounted in won't be updated when they change on the host. This may not be an issue here, or I could be remembering incorrectly, but is worth checking!

ereslibre:
> From memory, this means the content of the files mounted in won't be updated when they change on the host. This may not be an issue here, or I could be remembering incorrectly, but is worth checking!

Just checked it, had to remount `/` as `RW` on the MicroOS host to check this, and updating for example `/usr/share/salt/kubernetes/config/master.d/api.conf` did have instant effect on the mount `/etc/salt/master.d/api.conf`.

kiall:
Okay, cool - thanks for checking. I must be misremembering :)

PR #58, opened 2017-06-13
Connect the `salt-minion` in the administration dashboard machine to the `salt-master`

ereslibre:
Set the `admin` role to the administration dashboard machine, as well
as the minion configuration (`id` and `master` location).

This way we will leave the `salt-minion` in the administration dashboard
connected to the `salt-master` for future orchestrated upgrades.

### DIFF
--- a/activate.sh
+++ b/activate.sh
@@ -46,3 +46,14 @@ if ! [ -f /root/.ssh/id_rsa ]; then
 fi
 [ -d /var/lib/misc/ssh-public-key  ] || mkdir -p /var/lib/misc/ssh-public-key
 cp /root/.ssh/id_rsa.pub /var/lib/misc/ssh-public-key
+
+# Connect the salt-minion running in the administration controller node to the local salt-master
+# instance that is running in a container
+cat <<EOF > /etc/salt/grains
+roles:
+- admin
+EOF
+echo "master: localhost" > /etc/salt/minion.d/minion.conf
+echo "id: admin" > /etc/salt/minion.d/minion_id.conf
+
+systemctl enable salt-minion


mssola:
While testing kubic-project/terraform#72 I saw that at least in the latest image there's a certain `/etc/salt/minion.d/minion_id.conf` file that takes precedence over the ID set in `/etc/salt/minion.d/minion.conf`. Take that into account 😉

ereslibre:
@mssola Thanks for the heads up. The `/etc/salt/minion.d/minion_id.conf` is a cache of the `minion_id`. Setting `id:` in `minion.conf` has priority over anything else.

mssola:
> Setting id: in minion.conf has priority over anything else.

Not what I experienced 😕 What I've seen while testing that PR is that `/etc/salt/minion.d/minion_id.conf` takes precedence, regardless if you have `id:` in `minion.conf`. I checked this by calling `salt-key` on the master and realizing that the value comes from `minion_id.conf`, and not `minion.conf`.

ereslibre:
@mssola https://github.com/saltstack/salt/blob/4dc110a43d597ba13509f52b9da49d29765af0eb/salt/config/__init__.py#L3403-L3411

The workflow is always to honor `id:`, if not... then `get_id`, which will read the cache (`/etc/salt/minion.d/minion_id.conf`) -- if `minion_id_caching` is enabled, and if the cache is not present, it will try to guess the name.

If you explicit `id:` on the minion config, you never get to `get_id`.

grahamhayes:
Is the `salt-minion` service enabled by default on the admin node? If not, we need to also enable the service

ereslibre:
@grahamhayes That's what I wanted to verify. When doing so I will remove the `WIP` flag, thank you.

grahamhayes:
@ereslibre just checked on a test env - the `salt-minion` service is disabled by default

ereslibre:
> @ereslibre just checked on a test env - the salt-minion service is disabled by default

Thank you for checking. Now I need to verify if it's possible at all to `systemctl enable` inside the `activate.sh` script, or we will need to go with other solution.

Will report back ASAP.

ereslibre:
I was able to build an ISO with this package containing the new contents of the `activate.sh` script and I can confirm the following after a dashboard installation using this ISO:

```
linux-ne1k:~ # ps aux | grep salt-minion
root      1557  0.2  1.7 130136 17580 ?        Ss   16:49   0:00 /usr/bin/python /usr/bin/salt-minion
root      1609  0.9  3.5 378364 35924 ?        Sl   16:49   0:00 /usr/bin/python /usr/bin/salt-minion
root      1667  0.0  2.1 256776 21860 ?        S    16:49   0:00 /usr/bin/python /usr/bin/salt-minion
root      2199  0.0  0.1  10544  1576 pts/0    S+   16:50   0:00 grep salt-minion
```

```
linux-ne1k:~ # cat /etc/salt/minion.d/minion_id.conf 
id: admin
```

```
linux-ne1k:~ # cat /etc/salt/minion.d/minion.conf 
master: localhost
```

```
linux-ne1k:~ # cat /etc/salt/grains 
roles:
- admin
```

```
linux-ne1k:~ # salt-call grains.get osfullname
[WARNING ] Unable to find IPv6 record for "linux-ne1k" causing a 10 second timeout when rendering grains. Set the dns or /etc/hosts for IPv6 to clear this.
local:
    CAASP
```

```
linux-ne1k:~ # docker exec -it $(docker ps | grep salt-master | awk '{print $1}') salt '*' grains.get fqdn
admin:
    linux-ne1k
ca:
    linux-ne1k
```

Thus, the administration dashboard correctly registers against the salt-master service running in a container in this host.

activate.sh
```diff
@@ -46,3 +46,14 @@ if ! [ -f /root/.ssh/id_rsa ]; then
...
+# Connect the salt-minion running in the administration controller node to the local salt-master
+# instance that is running in a container
+cat <<EOF > /etc/salt/grains
+roles:
+- admin
+EOF
+echo "master: localhost" > /etc/salt/minion.d/minion.conf
+echo "id: admin" > /etc/salt/minion.d/minion_id.conf
+
+systemctl enable salt-minion
```
flavio: This is executed on the 1st run of the admin node, isn't it? If that's the case shouldn't we use `systemctl enable --now salt-minion` to ensure the process is also started immediately?

ereslibre: No, this is executed at the very end of the YaST installation, we don't have a `salt-master` yet running. After this script is executed, the machine will get rebooted.

flavio: I see, then the current code is completely fine.

kiall:
So, do we need to remove anything that targets '*' from our salt states before merging this? Then we can selectively enable things we want to run on the admin node.

ereslibre:
> So, do we need to remove anything that targets '*' from our salt states before merging this? Then we can selectively enable things we want to run on the admin node.

As far as I can tell all the commands that we are triggering with `tgt: '*'` are harmless (they only do salt related work, like sync grains, update mine...). I don't think it's necessary to change them before merging this, but let me do something. I am going to create a cluster with our salt states to confirm that everything worked fine.

ereslibre:
Can confirm I have created a cluster (1 master and 1 worker) with the admin registered and everything is  working as expected. The orchestration did fail at the very end:

```
----------
          ID: set_bootstrap_grain
    Function: salt.function
        Name: grains.setval
      Result: False
     Comment: 
     Started: 10:48:48.174866
    Duration: 138.649 ms
     Changes:   

Summary for dashboard_master
-------------
Succeeded: 15 (changed=3)
Failed:     1
-------------
```

No reason specified, I don't think this patch is related because `set_bootstrap_grain` is targeted to `roles:kube-(master|minion)`, but certainly there's something weird here. The current salt version:

```
dashboard:~ # salt-call --version
salt-call 2016.11.4 (Carbon)
```

ereslibre:
To make things even weirder, I can confirm I can set the grains manually:

```
dashboard:~ # docker exec -it $(docker ps | grep salt-master | awk '{print $1}') salt -C 'P@roles:kube-(master|minion)' grains.setval bootstrap_complete true
084816ac91a84b8cab2563dcec83523b:
    ----------
    bootstrap_complete:
        True
e8d63e7c427a4b29be4b288ddf4f3c8e:
    ----------
    bootstrap_complete:
        True
```

And retrieve them:

```
dashboard:~ # docker exec -it $(docker ps | grep salt-master | awk '{print $1}') salt -C 'P@roles:kube-(master|minion)' grains.get bootstrap_complete
e8d63e7c427a4b29be4b288ddf4f3c8e:
    True
084816ac91a84b8cab2563dcec83523b:
    True
```

ereslibre:
Okay, so https://github.com/kubic-project/salt/blob/14c58dccf3f20882cb68323e01b6115e927be44d/salt/orch/kubernetes.sls#L130-L138 is missing `tgt_type: grain_pcre`... how was this even working before?

ereslibre:
I can confirm setting `tgt_type: grain_pcre` did fix the orchestration issue and the orchestration is completely green.

ereslibre:
Opened PR for the issue https://github.com/kubic-project/salt/pull/120

ereslibre:
Another approval is needed for this code to be merged.

PR #62, opened 2017-06-22
bsc#1043592 Add pre-generation of minion keys

grahamhayes:
Generates 2 salt keys (ca and admin) and places them in the correct
directories.

This allows us to remove "auto_accept" from the master config file
and select the rest of the members of the cluster.

The admin key is writen out to `/etc/salt/pki/minion/minion.(pub|pem)`
The ca key is written out the same path in the container.

bsc#1043592

Current Issues:
- [ ] When using this in the devenv, salt calls take a long time, as this causes the salt master to try and wait for a non existent admin node.

### DIFF
--- a/public.yaml
+++ b/public.yaml
@@ -61,6 +61,39 @@ metadata:
         ]
       },
       {
+        "name": "salt-minion-key-generation",
+        "image": "sles12/salt-master:2016.11.4",
+        "command": ["sh", "-c", "umask 377;
+                                 mkdir /salt-master-pki/minions/;
+                                 if [ ! -f /salt-admin-minion-pki/minion.pub ] || [ ! -f /salt-admin-minion-pki/minion.pem ]; then
+                                   salt-key --gen-keys=admin --gen-keys-dir /tmp;
+                                   cp /tmp/admin.pub /salt-master-pki/minions/admin;
+                                   mv /tmp/admin.pub /salt-admin-minion-pki/minion.pub;
+                                   mv /tmp/admin.pem /salt-admin-minion-pki/minion.pem;
+                                 fi;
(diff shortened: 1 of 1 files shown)


public.yaml
```diff
@@ -61,6 +61,35 @@ metadata:
         ]
       },
       {
+        "name": "salt-minion-key-generation",
+        "image": "sles12/salt-master:2016.11.4",
+        "command": ["sh", "-c", "umask 377;
+                                 salt-key --gen-keys=ca --gen-keys-dir /tmp;
+                                 salt-key --gen-keys=admin --gen-keys-dir /tmp;
```
ereslibre: What happens after a restart? Should we only separate those in two blocks depending on each one for more safety?

```
if [ ! -f /salt-admin-minion-pki/minion.pub ] || [ ! -f /salt-admin-minion-pki/minion.pem ]; then
  salt-key --gen-keys=admin --gen-keys-dir /tmp;
  cp /tmp/admin.pub /salt-master-pki/minions/admin;
  mv /tmp/admin.pub /salt-admin-minion-pki/minion.pub;
  mv /tmp/admin.pem /salt-admin-minion-pki/minion.pem;
fi
if [ ! -f /salt-ca-minion-pki/minion.pub ] || [ ! -f /tmp/ca.pem /salt-ca-minion-pki/minion.pem ]; then
  cp /tmp/ca.pub /salt-master-pki/minions/ca;
  salt-key --gen-keys=ca --gen-keys-dir /tmp;
  mv /tmp/ca.pub /salt-ca-minion-pki/minion.pub;
  mv /tmp/ca.pem /salt-ca-minion-pki/minion.pem;
fi
```

grahamhayes: Yeah - had not considered this - I **think** salt would still work, but it is better to be safe. Will add ^

ereslibre:
Green tests, two approvals, merging despite of the admin issue -- we already have a workaround that we will soon include.

PR #63, opened 2017-06-23
Add admin node setup service

jordimassaguerpla:


jordimassaguerpla:
package for testing:
https://build.suse.de/package/show/home:jordimassaguerpla:branches:Devel:CASP:1.0:ControllerNode/caasp-container-manifests

jordimassaguerpla:
I've tested this on a sles12sp2 vm because I was not able to build an iso locally with this package.

activate.sh
```diff
@@ -58,3 +36,4 @@ echo "id: admin" > /etc/salt/minion.d/minion_id.conf
 echo "grains_refresh_every: 10" > /etc/salt/minion.d/grains_refresh.conf
 
 systemctl enable salt-minion
+systemctl enable admin-node-setup
```
kiall: This won't work right on cloud-init based deployments (and the salt-minion one is due to be removed). This needs to be added to YaST and our cloud-init module.

That said, we should leave it here for now - because we have an ordering problem to get all 3 changes landed! Leaving it here will mean we don't break things while we await the other 2 changes.

jordimassaguerpla:
I tried installing the package in a CaaSP host. Running the service works fine. Which I couldn't test was the "firstboot" as I couldn't build an iso myself. Could we validate that with the Devel iso?

PR #65, opened 2017-06-27
bsc#1043592 Use mktemp to create tmp directories

grahamhayes:
Use `mktemp` to ensure that directory has a random name

### DIFF
--- a/public.yaml
+++ b/public.yaml
@@ -65,18 +65,20 @@ metadata:
         "image": "sles12/salt-master:2016.11.4",
         "command": ["sh", "-c", "umask 377;
                                  mkdir /salt-master-pki/minions/;
+                                 temp_dir=`mktemp -d`;
                                  if [ ! -f /salt-admin-minion-pki/minion.pub ] || [ ! -f /salt-admin-minion-pki/minion.pem ]; then
-                                   salt-key --gen-keys=admin --gen-keys-dir /tmp;
-                                   cp /tmp/admin.pub /salt-master-pki/minions/admin;
-                                   mv /tmp/admin.pub /salt-admin-minion-pki/minion.pub;
-                                   mv /tmp/admin.pem /salt-admin-minion-pki/minion.pem;
+                                   salt-key -u root --gen-keys=admin --gen-keys-dir $temp_dir;
+                                   cp $temp_dir/admin.pub /salt-master-pki/minions/admin;
+                                   mv $temp_dir/admin.pub /salt-admin-minion-pki/minion.pub;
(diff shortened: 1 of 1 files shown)


public.yaml
```diff
@@ -65,17 +65,18 @@ metadata:
         "image": "sles12/salt-master:2016.11.4",
         "command": ["sh", "-c", "umask 377;
                                  mkdir /salt-master-pki/minions/;
+                                 temp_dir=`mktemp -d`;
                                  if [ ! -f /salt-admin-minion-pki/minion.pub ] || [ ! -f /salt-admin-minion-pki/minion.pem ]; then
-                                   salt-key --gen-keys=admin --gen-keys-dir /tmp;
-                                   cp /tmp/admin.pub /salt-master-pki/minions/admin;
-                                   mv /tmp/admin.pub /salt-admin-minion-pki/minion.pub;
-                                   mv /tmp/admin.pem /salt-admin-minion-pki/minion.pem;
+                                   salt-key -u root --gen-keys=admin --gen-keys-dir $temp_dir;
```
kiall: I assume the addition of `-u root` is for forcing salt to match the perms of the mktemp created directly?

public.yaml
```diff
@@ -65,17 +65,18 @@ metadata:
...
-                                   salt-key --gen-keys=ca --gen-keys-dir /tmp;
-                                   cp /tmp/ca.pub /salt-master-pki/minions/ca;
-                                   mv /tmp/ca.pub /salt-ca-minion-pki/minion.pub;
-                                   mv /tmp/ca.pem /salt-ca-minion-pki/minion.pem;
+                                   salt-key -u root --gen-keys=ca --gen-keys-dir $temp_dir;
+                                   cp $temp_dir/ca.pub /salt-master-pki/minions/ca;
+                                   mv $temp_dir/ca.pub /salt-ca-minion-pki/minion.pub;
+                                   mv $temp_dir/ca.pem /salt-ca-minion-pki/minion.pem;
                                  fi;
                                  exit 0"],
```
kiall: <darix> kiall: generally good. i would recommend cleaning up the temp_dir at the end.

Lets `rm -rf` the directory before exit 0.

public.yaml
grahamhayes: Yup - mktemp is ran as `root` so the directory is root owned, and salt by default runs as the salt user, `-u root`

public.yaml
grahamhayes: the container is deleted after the `exit 0` ... which by default deletes the directory ...

grahamhayes: Done

PR #66, opened 2017-06-28
bsc#1045350 Accept salt keys that have been pre-generated

grahamhayes:
Currently the admin nodes `salt-minion` starts before the container
that generates and accepts keys is ran.

This means that the salt minion is started with a key that is not
accepted, and goes to a pending state.

This checks if the key is pre-generated, and if we have accepted
a key from this minion before. If the key has been generated,
but not accepted, we accept the key and continue.

### DIFF
--- a/public.yaml
+++ b/public.yaml
@@ -66,11 +66,17 @@ metadata:
         "command": ["sh", "-c", "umask 377;
                                  mkdir /salt-master-pki/minions/;
                                  temp_dir=`mktemp -d`;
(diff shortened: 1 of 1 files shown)


public.yaml
```diff
@@ -66,12 +66,18 @@ metadata:
         "command": ["sh", "-c", "umask 377;
                                  mkdir /salt-master-pki/minions/;
                                  temp_dir=`mktemp -d`;
+                                 if [ -f /salt-admin-minion-pki/minion.pub ] && [ -f /salt-admin-minion-pki/minion.pem ] && [ ! -f /salt-master-pki/minions/admin ]; then
```
ereslibre: Can't this be expressed as the else branch of the `if [ ! -f /salt-admin-minion-pki/minion.pub ] || [ ! -f /salt-admin-minion-pki/minion.pem ]; then` ?

PR #69, opened 2017-06-28
Clear TX update grains on admin node boot bsc#1045379

kiall:
Clear the tx_update_{reboot_needed,failed} grains upon boot. This ensures
the UI doesn't continue to show an admin node upgrade after we've upgraded.

### DIFF
--- a/admin-node-setup.sh
+++ b/admin-node-setup.sh
@@ -1,9 +1,15 @@
 #!/bin/sh
 # this script WILL BE RUN ON EVERY REBOOT
 
+# Clear the transactional update grains when booting up
+if [ -f /etc/salt/grains ]; then
+    sed -i -e 's|tx_update_reboot_needed:.*|tx_update_reboot_needed: false|g' /etc/salt/grains
+fi
+
 # Update manifest files
 kube_dir=/etc/kubernetes/manifests
 manifest_dir=/usr/share/caasp-container-manifests
+
 if [ ! -d $kube_dir ]; then
     echo "$kube_dir does not exist" >&2
     echo "make sure kubernetes is installed" >&2
@@ -22,7 +28,6 @@ if [ ! -f $manifest_dir/private.yaml ]; then
     echo "private.yaml is not in $manifest_dir" >&2
     exit -3
 fi
+
 cp -v $manifest_dir/public.yaml $kube_dir
 cp -v $manifest_dir/private.yaml $kube_dir
-
-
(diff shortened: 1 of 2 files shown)


kiall:
Tested by creating a drop in unit on an admin node /etc/systemd/system/admin-node-setup.service.d/01-test.conf:

	[Unit]
	Before=kubelet.service salt-minion.service

	[Service]
	ExecStart=
	ExecStart=/etc/admin-node-setup.sh

then putting the updated script into /etc/admin-node-setup.sh + chmod +x..

Then finally reboot, and ensure the tx_update grains grains have been reset to false

kiall:
Also just walked through an admin node upgrade with these changes, and the flag was correctly cleared from velum

jordimassaguerpla:
I created this package to test
http://download.suse.de/ibs/home:/jordimassaguerpla:/branch:/CAASP:/bsc_1045379/standard/noarch/

admin-node-setup.sh
```diff
@@ -1,9 +1,16 @@
 #!/bin/sh
 # this script WILL BE RUN ON EVERY REBOOT
 
+# Clear the transactional update grains when booting up
+if [ -f /etc/salt/grains ]; then
+    sed -i -e 's|tx_update_reboot_needed:.*|tx_update_reboot_needed: false|g' /etc/salt/grains
+    sed -i -e 's|tx_update_failed:.*|tx_update_failed: false|g' /etc/salt/grains
```
grahamhayes: should we clear this on a reboot? This gets set when the snapshot creation failed, so even after a reboot, it is still in a failed state.

kiall: That's a good point. TX update will manage the lifecycle of the failed grain.

I'll remove it.

kiall:
Copying over from IRC:

<kiall> jordi: it seems TX update decided it preferred the not installed by hand RPM??
<kiall> It got downgraded
<kiall>     The following package is going to be downgraded:
<kiall>       caasp-container-manifests
<kiall>     
<kiall> I have no understanding of why it would have done that

admin-node-setup.sh
```diff
@@ -1,9 +1,15 @@
 #!/bin/sh
 # this script WILL BE RUN ON EVERY REBOOT
 
+# Clear the transactional update grains when booting up
+if [ -f /etc/salt/grains ]; then
+    sed -i -e 's|tx_update_reboot_needed:.*|tx_update_reboot_needed: false|g' /etc/salt/grains
```
inercia: A `sed -i "/tx_update_reboot_needed/d" /etc/salt/grains` would just remove the line

kiall: Is there any practical reason to prefer removal over setting it back to false? The decision to reset it to false rather than true was based on mimicking the logic in transactional-update itself.

PR #72, opened 2017-06-30
Reuse `$DIR` when setting the prefix for the private folder

ereslibre:
Move constant functions to global vars.

### DIFF
--- a/gen-certs.sh
+++ b/gen-certs.sh
@@ -9,47 +9,36 @@ CITY=${CITY:-Nuremberg}
 STATE=${STATE:-Bavaria}
 COUNTRY=${COUNTRY:-DE}
 
-dir() {
-    echo "/etc/pki"
-}
-
-certs() {
-    echo "$(dir)/_certs"
-}
-
-privatedir() {
-    echo "/etc/pki/private"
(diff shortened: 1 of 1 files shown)


gen-certs.sh
```diff
@@ -18,7 +18,7 @@ certs() {
 }
 
 privatedir() {
-    echo "/etc/pki/private"
+    echo "$(dir)/private"
```
inercia: Wouldn't it be better to replace `$(dir)` (a function call) by a simple variable?

flavio: Yes, this looks a bit odd to me too...

ereslibre: Yes, this can be changed to a variable without much hassle (this and other functions).

This PR doesn't need to be part of the release, so I'll tidy it up afterwards.

ereslibre:
[Reuse `$(dir)` when setting the prefix for the private folder](https://trello.com/c/k4BcZBX4/454-reuse-dir-when-setting-the-prefix-for-the-private-folder)

ereslibre:
Rebase done, and comments have been taken care of.

PR #81, opened 2017-07-17
set workers to 1 for the autoyast container

MaximilianMeister:
this instance of velum will likely not handle many concurrent
requests

## DEPENDS ON
https://github.com/kubic-project/velum/pull/257

### DIFF
--- a/public.yaml
+++ b/public.yaml
@@ -283,6 +283,8 @@ spec:
       value: /var/lib/misc/infra-secrets/mariadb-velum-password
     - name: VELUM_DB_SOCKET
       value: /var/run/mysql/mysql.sock
+    - name: VELUM_WORKERS
+      value: "1"
     volumeMounts:
       - mountPath: /var/run/mysql
         name: mariadb-unix-socket


MaximilianMeister:
[Bug 1048533 - Velum connection limit leading to lockups](https://trello.com/c/AJP5EZ8R/290-bug-1048533-velum-connection-limit-leading-to-lockups)

kiall:
I'm not sure if we're likely to get lots of autoyast requests at once, which might trigger this issue on this container too? e.g. booting 50 nodes, where YaST requests the same XML file 3-4 times for each node. I'm not sure we should put this into 1.0.

ereslibre:
@kiall If you don't feel confident with this change I'm okay with postponing it, so we can properly test it. My guts tell me we will serve this without any trouble, but it's safer to validate it with bigger clusters, that's for sure.

kiall:
@ereslibre I think it's fine to merge to master, I'm just not sure we should put it into release-1.0 - given there is no blocking bug it's fixing.

ereslibre:
Totally agree, sorry didn't understand. I'd not backport this patch to 1.0 as it's not fixing any explicit bug.

kiall:
Actually, interestingly - Jenkins CI failed on it waiting for velum to start up!

ereslibre:
Yeah, but only one of those. I was going to ask for a retrigger but I see it's already building :)

MaximilianMeister:
> Actually, interestingly - Jenkins CI failed on it waiting for velum to start up!

I also had troubles running that code locally, kubelet fails with a parsing error from the manifest... i'll investigate more later, so lets put this on hold

ereslibre:
> I also had troubles running that code locally, kubelet fails with a parsing error from the manifest... i'll investigate more later, so lets put this on hold

Oh, try with `value: "1"`, it doesn't like integers as envvar values.

MaximilianMeister:
> Oh, try with value: "1", it doesn't like integers as envvar values.

fixed, thanks for the hint :+1:

MaximilianMeister:
> I think it's fine to merge to master, I'm just not sure we should put it into release-1.0 - given there is no blocking bug it's fixing.
> 
> In master, it can get some time to bake and issues to show up :)

I'd rather think that issues could pop up if we don't backport it. I would still consider to backport it, as it reduces the memory consumption, but i agree it's not critical now.

@kiall could we keep the `release-1.0` label to not lose track of it? we could backport it also after the release as a maintenance update

kiall:
> I'd rather think that issues could pop up if we don't backport it. I would still consider to backport it, as it reduces the memory consumption, but i agree it's not critical now.

What issues do you think are likely? The only thing I can think of is a little more memory consumption, however, that seems OK to me as we know Puma fails with more than 1 connection and only 1 worker.

Anyone deploying a reasonable number of machines with autoyast will end up having more than 1 machine talk to this container at once, potentially leading to the same lockup we increased the worker count to 2 in avoid?

MaximilianMeister:
> Anyone deploying a reasonable number of machines with autoyast will end up having more than 1 machine talk to this container at once, potentially leading to the same lockup we increased the worker count to 2 in order to avoid?

hmm ok, haven't tested this but if that would create a similar lockup, it sounds like a reason to keep the same worker number. let's close this for now, we can come back to it later if we run into issues

PR #82, opened 2017-07-18
Add fingerprints to the velum issue

ereslibre:
By adding SHA1 and SHA256 fingerprints to the Velum issue, we can
ensure that the instance we are accessing is the right one, and we
are not mistaken (several clusters) or to reject a MITM, since the
certificates chain of trust does not exist (the CA is autogenerated),
and the customer has no way to import the CA as trusted for now.

Fixes: bsc#1048135


ereslibre:
I built an ISO using `build-caasp-iso` project with this patch applied, and can confirm on the issue the fingerprints of the velum certificate is shown:

screenshot from 2017-07-18 17-40-11

What matches in this specific instance with what was reported by my browser:

screenshot from 2017-07-18 17-44-00

flavio:
Note well: @ukbelch approved that over bugzilla

PR #90, opened 2017-08-11
etcd on admin node does not have any peers

kiall:
As such, there is no reason to listen on 0.0.0.0 for
peering.

With etcd 3.2.4, this configuration actually does not
work, giving an error:

    Aug 11 15:15:55 caasp-admin etcd[2688]: --initial-cluster must include default=http://10.17.1.0:2380 given --initial-advertise-peer-urls=http://10.17.1.0:2380

Since we don't actually peer, the simplest solution is
to just not half configure peering.

### DIFF
--- a/activate.sh
+++ b/activate.sh
@@ -9,7 +9,6 @@ if ! grep "pod-infra-container-image" /etc/kubernetes/kubelet &> /dev/null; then
 fi
 
 # Make sure etcd listens on 0.0.0.0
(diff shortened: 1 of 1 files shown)


kiall:
Tagged with release-1.0-candidate as, any port left open that is not needed should generally be closed for security reasons. A 1.0 backport should be harmless.

kiall:
Validated that the admin node now correctly boots up and etcd starts:

    caasp-admin:~ # systemctl status etcd
    ● etcd.service - Etcd Server
       Loaded: loaded (/usr/lib/systemd/system/etcd.service; enabled; vendor preset: disabled)
       Active: active (running)

PR #91, opened 2017-08-17
Increase MariaDBs max_allowed_packet to 16MB

kiall:
MariaDB's max allowed packet size is too small for some larger
deployments, by increasing it, we allow ourselves some time to
implement an alternative pattern for handling salt's event
stream.

Fixes bsc#1054250

### DIFF
--- a/config/mysql/max-allowed-packet.cnf
+++ b/config/mysql/max-allowed-packet.cnf
@@ -1,0 +1,2 @@
+[mysqld]
+max_allowed_packet=16M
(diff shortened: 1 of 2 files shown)


kiall:
Any objections? WIP as I'm still validating 16MB is enough

grahamhayes:
I am booting a stack to check right now

kiall:
Between this, and disabling job_cache, we're back to successfully deploying 50 nodes.

PR #92, opened 2017-08-17
Include haproxy as a dependency

kiall:
haproxy will be used to loadbalance requests over the chosen masters, so
we'll need to include the haproxy docker RPM as a dependency.


kiall:
So - WIP for 2 reasons..

1) The haproxy image isn't showing up in http://download.suse.de/ibs/Devel:/CASP:/Head:/ControllerNode/standard/x86_64/ - I'm not sure why

2) Adding this a dependency of the container manifests feels wrong, however, that's where ALL our container images have dependencys...

kiall:
Unmarking as WIP, the haproxy container RPM now exists, and this is where we specify these for better/worse... If we want to change change that, it's unrelated to adding haproxy.

PR #94, opened 2017-08-22
Use the tagfiles rather than hardcoding a tag

kiall:
This decouples our manifests from the tag of the images contained
within the RPMs. These tagfiles will contain the latest, and most
specifc, tag for a given image.

Depends-On: https://github.com/SUSE/containment-rpm-docker/pull/23

### DIFF
--- a/admin-node-setup.sh
+++ b/admin-node-setup.sh
@@ -9,6 +9,7 @@ fi
 # Update manifest files
 kube_dir=/etc/kubernetes/manifests
 manifest_dir=/usr/share/caasp-container-manifests
+images_dir=/usr/share/suse-docker-images/native
 
 if [ ! -d $kube_dir ]; then
     echo "$kube_dir does not exist" >&2
@@ -29,5 +30,21 @@ if [ ! -f $manifest_dir/private.yaml ]; then
     exit -3
 fi
 
-cp -v $manifest_dir/public.yaml $kube_dir
-cp -v $manifest_dir/private.yaml $kube_dir
+tmp_dir=$(mktemp -d)
+
+cp $manifest_dir/public.yaml $tmp_dir
+cp $manifest_dir/private.yaml $tmp_dir
+
+for i in $(ls $images_dir/sles*.tag);do
+    metadata_file=$(basename $i .tag).metadata
+    image_name=$(cat $images_dir/$metadata_file | grep \"name\": | cut -d":" -f2 | cut -d\" -f2)
+    tag=$(cat $i)
+    echo "$0: Setting $image_name:$tag into public and private manifests"
+    sed -i -e "s%$image_name:__TAG__%$image_name:$tag%g" $tmp_dir/public.yaml
+    sed -i -e "s%$image_name:__TAG__%$image_name:$tag%g" $tmp_dir/private.yaml
+done
+echo "$0: Replacing public and private manifests"
+cp $tmp_dir/public.yaml $kube_dir
+cp $tmp_dir/private.yaml $kube_dir
+
+rm -rf $tmp_dir
--- a/private.yaml
+++ b/private.yaml
@@ -21,7 +21,7 @@ metadata:
     pod.beta.kubernetes.io/init-containers: '[
       {
         "name": "mariadb-secrets",
-        "image": "sles12/velum:0.0",
+        "image": "sles12/velum:__TAG__",
         "command": ["sh", "-c", "umask 377;
                                  if [ ! -f /infra-secrets/mariadb-root-password ]; then
                                    head -n 10 /dev/random | base64 | head -n 10 | tail -n 1 > /infra-secrets/mariadb-root-password;
@@ -39,7 +39,7 @@ spec:
   hostNetwork: false
   containers:
   - name: velum-mariadb
-    image: sles12/mariadb:10.0
+    image: sles12/mariadb:__TAG__
     env:
     - name: MYSQL_DISABLE_REMOTE_ROOT
       value: "true"
--- a/public.yaml
+++ b/public.yaml
@@ -21,7 +21,7 @@ metadata:
     pod.beta.kubernetes.io/init-containers: '[
       {
         "name": "mariadb-user-secrets",
-        "image": "sles12/mariadb:10.0",
+        "image": "sles12/mariadb:__TAG__",
         "command": ["/setup-mysql.sh"],
         "env": [
           {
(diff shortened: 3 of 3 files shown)


kiall:
This will fail CI until https://github.com/SUSE/containment-rpm-docker/pull/23 is merged, and RPMs are rebuilt using the new containment-rpm-docker.

ereslibre:
This also has an impact on https://github.com/kubic-project/caasp-devenv, as it will need to make this substitutions on the manifests.

Just a heads up that it needs to be considered.

kiall:
@ereslibre caasp-devenv should already be calling this script, so it should just work.

(Worth mentioning that the "newer" tooling (caasp-kvm) certainly does call this, rather than manually hacking these steps in)

ereslibre:
> @ereslibre caasp-devenv should already be calling this script, so it should just work.

AFAIK this isn't the case right now as the paths specified in this script are sometimes absolute and would not work with the caasp-devenv. In any case this script is not called from the devenv at the moment.

> (Worth mentioning that the "newer" tooling (caasp-kvm) certainly does call this, rather than manually hacking these steps in)

Yup, this is the case. I was mentioning the problem with caasp-devenv as I still find it useful for the general use case when developing (note that I'm not against other methods or to completely dismiss caasp-devenv once that we have a better environment that has all the developer oriented stuff it has).

kiall:
Oh - I just checked, thought we fixed that in caasp-devenv.. 

Once https://github.com/kubic-project/container-feeder/pull/26 lands - the same replacements could be done, however dropping in "latest" instead of the content of the tag file? Maybe you could work up a fix @ereslibre ? I won't have time myself :(

ereslibre:
Of course, I'll take care of it. Thank you :)

jordimassaguerpla:
I think this could be improved by, instead of using specific tags for each container, we would replace the whole "IMAGE_NAME:TAG", so we could do a loop. Something like:

cp /usr/... /etc/....
for i in $(ls /usr/share/suse-docker-images/native/*.tag);do
    # remove .tag from image name and append .metadata to get the metadata file
    # get image_name from metadata file
    tag=$(cat /usr/share/suse-docker-images/native/$i)
    sed -e "s/$image_name:TAG/$image_name:$tag/g" -i /etc/....
done

kiall:
FWIW, we cant / shouldn't copy it into /etc/k8s/manifests BEFORE editing the files, and Kubernetes will immediately start trying to boot the containers. The files in /etc/k8s/manifests should never be an intermediate version.

Is there a reason we need to replace the whole image name? Those are reasonably fixed (until sles15) I'm not sure it buys us much?

kiall:
Also - the tag file works out to something like this, once all changes have landed:

     /usr/share/suse-docker-images/native/sles12-mariadb.x86_64-1.0.4.metadata 

to get the image name, we would need to parse out `sles12-mariadb` and swap the - for a /, giving us `sles12/mariadb`.. However, this will be more awkward with e.g. `sles12-salt-master` - which needs to end up as `sles12/salt-master`

And - it gets even more awkward when we update our image names to match those which will be used on the SUSE container registry, which will be something like `caasp/sles12/salt-master`

jordimassaguerpla:
The image name is *in* the metadata file, as:

```
cat ./usr/share/suse-docker-images/native/sles12-mariadb.x86_64-1.0.4.metadata
{
  "image": {
    "name": "sles12/mariadb",
    "tags": [ "10.0", "latest" ],
    "file": "sles12-mariadb.x86_64-1.0.4.tar.xz"
  }
}
```

kiall:
Ahh, I missed that :)

Would you mind updating? keeping in mind, we have to avoid editing inside /etc/k8s/manifests ?

jordimassaguerpla:
Can I copy the file to a temp location and edit there?

kiall:
Yea, it can be copied to another location, edited, and moved to the final /etc/k8s/manifests location

jordimassaguerpla:
ok. I can take care of this.

jordimassaguerpla:
@ereslibre : I refactored kiall's code to make it more generic. Could you review? I know is bash-magic code ... but still I think it can be right :)

admin-node-setup.sh
```diff
@@ -29,5 +30,21 @@ if [ ! -f $manifest_dir/private.yaml ]; then
...
+    metadata_file=$(basename $i .tag).metadata
+    image_name=$(cat $images_dir/$metadata_file | grep \"name\": | cut -d":" -f2 | cut -d\" -f2)
+    tag=$(cat $i)
+    echo "$0: Setting $image_name:$tag into public and private manifests"
+    sed -i -e "s%$image_name:__TAG__%$image_name:$tag%g" $tmp_dir/public.yaml
+    sed -i -e "s%$image_name:__TAG__%$image_name:$tag%g" $tmp_dir/private.yaml
+done
+echo "$0: Replacing public and private manifests"
+cp $tmp_dir/public.yaml $kube_dir
+cp $tmp_dir/private.yaml $kube_dir
```
ereslibre: Nitpick: you can `mv` the final files and `rmdir $tmp_dir` instead of `rm -f $tmp_dir`.

jordimassaguerpla: :) . It is a personal taste of mine ... I don't like moving files but prefer to copy and then remove them. I once lost all of my university assignments by moving files around...

ereslibre: Moving files does minimal operations on FS (inode of the file is kept for example), while copying a file for later destroying the original will perform more operations, but I guess we don't want to get that nitpicky here :)

ereslibre:
Related PR: https://github.com/kubic-project/caasp-devenv/pull/37

Ideally https://github.com/kubic-project/caasp-devenv/pull/37 should be merged before this one, because that change is not intrusive (everything will work as before), only that it will also replace `__TAG__` when found in a manifest.

mjura:
[Node Removal: Credentials no longer needed - e.g. TLS keys or passwords. How to invalidate?](https://trello.com/c/IiRAxAgV/556-node-removal-credentials-no-longer-needed-eg-tls-keys-or-passwords-how-to-invalidate)

mjura:
[Node Removal: Credentials no longer needed - e.g. TLS keys or passwords. How to invalidate?](https://trello.com/c/IiRAxAgV/556-node-removal-credentials-no-longer-needed-eg-tls-keys-or-passwords-how-to-invalidate)

jordimassaguerpla:
@mjura ?

PR #96, opened 2017-08-22
disable haproxy and flannel dependencies temporarily

jordimassaguerpla:
They can not be added until they are in the iso. Once they are in the
iso, we can enable them back


jordimassaguerpla:
@inercia : Will disabling these packages break something else?

jordimassaguerpla:
If we do that, we are going to break haproxy and block development for flannel-cni. Let's wait for the packages to be available.

PR #98, opened 2017-08-31
update requirements of docker images

jordimassaguerpla:
admin-node-setup.sh script expects images to have a .tag file in
order to substitute the __TAG__ tags in public.yaml and private.yaml

This .tag is in the update images, which are >= 1.1.0


jordimassaguerpla:
This can't be merged without updating the docker images first ... I would make all of them be 2.0.0 to make it explicit this are casp 2.0 plus they are seen as updates for the ones in casp1.0, which are 1.1.*

PR #102, opened 2017-09-06
Revert "update image requirements"

jordimassaguerpla:
This reverts commit 777e2226d8055566212bd7fc16e5b9324210fa0a.

This broke our dvd cause the new packages are not yet in. Let's revert
it and do this again once the new packages are in.


jordimassaguerpla:
Merging because this broke our iso.

PR #108, opened 2017-09-11
Include helm-tiller as a dependency

SlickNik:
The helm-tiller image can be optionally installed through velum, so
include the docker RPM for this as a dependency.


SlickNik:
Please do not merge until the SR for sles12sp3-helm-tiller-image is accepted. I will update with more info once that happens.

SlickNik:
This is no longer needed after https://github.com/kubic-project/caasp-container-manifests/commit/fed0ac3010695226d9a7b3212cd26c71d6be5233

Thanks Jordi! 👍

PR #111, opened 2017-09-14
Revert requirement on helm-tiller image.

jordimassaguerpla:
We will add this later.


jordimassaguerpla:
Adding WIP tag until a decision is final on if helm-tiller goes in now or later.

PR #113, opened 2017-09-21
Allow custom options to be passed to the Salt Master

kiall:
Create a file for custom salt-master configuration options
to be supplied. This will be loaded in numeric order, allowing
for certain options (e.g. worker thread counts).

bsc#1059724


kiall:
Force merging due to cyclic dep

PR #115, opened 2017-09-22
Follow prefix patterns for returner credentials too

ereslibre:
As has been introduced in 17a0d8d8ac58ee8cb6d79849219b5631a60afa1e

### DIFF
--- a/public.yaml
+++ b/public.yaml
@@ -89,10 +89,10 @@ spec:
   - name: salt-master-config
     image: sles12/velum:__TAG__
     command: ["sh", "-c", "umask 377;
-                           if ! grep mysql /salt-master-credentials/returner-credentials.conf; then
-                             echo \"mysql:\" > /salt-master-credentials/returner-credentials.conf;
-                             echo -n \"  pass: \" >> /salt-master-credentials/returner-credentials.conf;
-                             cat /infra-secrets/mariadb-salt-password >> /salt-master-credentials/returner-credentials.conf;
+                           if ! grep mysql /salt-master-credentials/55-returner-credentials.conf; then
+                             echo \"mysql:\" > /salt-master-credentials/55-returner-credentials.conf;
+                             echo -n \"  pass: \" >> /salt-master-credentials/55-returner-credentials.conf;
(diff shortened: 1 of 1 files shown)


flavio:
This fix look important to me. The code seems sane, but I didn't try it locally.

I triggered another run of the CI system.

public.yaml
```diff
@@ -120,11 +120,15 @@ metadata:
         "name": "salt-master-config",
         "image": "sles12/velum:__TAG__",
         "command": ["sh", "-c", "umask 377;
-                                 rmdir /salt-master-credentials/returner-credentials.conf;
-                                 if [ ! -f /salt-master-credentials/returner-credentials.conf ]; then
-                                   echo \"mysql:\" > /salt-master-credentials/returner-credentials.conf;
-                                   echo -n \"  pass: \" >> /salt-master-credentials/returner-credentials.conf;
-                                   cat /infra-secrets/mariadb-salt-password >> /salt-master-credentials/returner-credentials.conf;
+                                 rmdir /salt-master-credentials/55-returner-credentials.conf;
```
MaximilianMeister: doesn't `rmdir` fail when it's a file? but it has been there before, so there must be some reason?

ereslibre: This is exactly the point. If the file does not exist, the kubelet will `mkdir` it (because of the `mountPath`), so if it is a directory we `rmdir` it, if it's a file it will do anything and we just continue.

MaximilianMeister: so if the file is there when the code is run, the `rmdir` statement will fail, but won't fail hard?

monstermunchkin:
[Bug 1062248 — Naming convention is not unified in salt master configuration files](https://trello.com/c/tcTSrkKh/414-bug-1062248-naming-convention-is-not-unified-in-salt-master-configuration-files)

public.yaml
MaximilianMeister: in other words: `rmdir` on a file will always fail. is this a problem here?
```
% touch /tmp/somefile                                                                                                                                                                                      
% rmdir /tmp/somefile                                                                                                                                                                                            
rmdir: failed to remove '/tmp/somefile': Not a directory
% echo $?                                                                                                                                                                                                    1 ↵ ✭
1
```

ereslibre:
Rebased.

PR #120, opened 2017-10-04
Re-add the sle12-flannel-image

inercia:
Include [this package](https://build.suse.de/package/show/Devel:CASP:Head:ControllerNode/sles12sp3-flannel-image) in the CaaSP image.


jordimassaguerpla:
@inercia : Should this also go into 2.0?

inercia:
> @inercia : Should this also go into 2.0?

@jordimassaguerpla No, this is not for 2.0, only for HEAD.

PR #121, opened 2017-10-06
kubelet: switch deprecated --config flag to --kubeconfig

MaximilianMeister:
this needs to be done during upgrade to 2.0 otherwise kubelet
wont start

### DIFF
--- a/admin-node-setup.sh
+++ b/admin-node-setup.sh
@@ -48,3 +48,6 @@ cp $tmp_dir/public.yaml $kube_dir
 cp $tmp_dir/private.yaml $kube_dir
 
 rm -rf $tmp_dir
+
+# switch deprecated --config flag in kubelet
+sed -i -e "s/--config=/--pod-manifest-path=/g" /etc/kubernetes/kubelet


admin-node-setup.sh
```diff
@@ -48,3 +48,6 @@ cp $tmp_dir/public.yaml $kube_dir
 cp $tmp_dir/private.yaml $kube_dir
 
 rm -rf $tmp_dir
+
+# switch deprecated --config flag in kubelet
+sed -i -e "s/--config/--kubeconfig/g" /etc/kubernetes/kubelet
```
kiall: should be moved to `--pod-manifest-path`

MaximilianMeister: updated

admin-node-setup.sh
```diff
@@ -48,3 +48,6 @@ cp $tmp_dir/public.yaml $kube_dir
 cp $tmp_dir/private.yaml $kube_dir
 
 rm -rf $tmp_dir
+
+# switch deprecated --config flag in kubelet
+sed -i -e "s/--config/--pod-manifest-path/g" /etc/kubernetes/kubelet
```
ereslibre: Nit: I'd include `--config=` and `--pod-manifest-path=`. Just to make sure we don't replace anything starting with `--config`. It probably doesn't exist (don't have a cluster at hand to check), but just for sanity.

kiall:
https://bugzilla.suse.com/show_bug.cgi?id=1062011

admin-node-setup.sh
MaximilianMeister: done

MaximilianMeister:
fixed @ereslibre comment and added the bug reference

MaximilianMeister:
included in https://github.com/kubic-project/caasp-container-manifests/pull/123

PR #122, opened 2017-10-06
Move salt-master-custom.conf creation to admin-node-setup

kiall:
activate.sh is only ran once, during a fresh install. This step is required
on fresh installs, and 1.0 -> 2.0 upgrades, so moving to admin-node-setup.sh
and ensuring idempotency will resolve this issue.

bsc#1062003

### DIFF
--- a/activate.sh
+++ b/activate.sh
@@ -1,6 +1,7 @@
 #!/bin/sh
 # activate the dashboard
 # this script WILL BE RUN ONLY ONCE, after the installation
+# this script WILL NOT RUN during/after upgrades
 
 # Make sure that the controller node looks for the local pause image
 # TODO: remove this as soon as possible. As an idea, we could use a systemd drop-in unit.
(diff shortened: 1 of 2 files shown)


activate.sh
```diff
@@ -28,15 +29,6 @@ EOF
 echo "master: localhost" > /etc/salt/minion.d/minion.conf
```
inercia: Why don't we surround all this code with a

```bash
if [ ! -f /etc/sysconfig/caasp-activation ] ; then
   touch /etc/sysconfig/caasp-activation
   # the rest of the code here...
fi
```

so we are completely sure this is run only once after the installation?

admin-node-setup.sh
```diff
@@ -48,3 +48,12 @@ cp $tmp_dir/public.yaml $kube_dir
 cp $tmp_dir/private.yaml $kube_dir
 
 rm -rf $tmp_dir
+
+# First time setup of user-configuration for salt-master
+if [ ! -d "/etc/caasp" ]; then
+    mkdir /etc/caasp
+fi
```
inercia: This can be reduced to just a `mkdir -p /etc/caasp` (it does not fail if the directory already exists)

PR #128, opened 2017-10-09
Wait for network before running admin-node-setup.service

kiall:
As we need to know the IPs and hostnames for use in the TLS certificates
this generates, we should ensure the network is up and running before this
unit triggers.

bsc#1062284

Backport of https://github.com/kubic-project/caasp-container-manifests/pull/127


kiall:
CI is broken, and wouldn't test this particular change anyway. Merging based on the +A, and Rob K's validation of the fix within the environment he found it it.

PR #131, opened 2017-10-17
Export /etc/hosts from host to velum-dashboard pod, bsc#1062728

mjura:
Docker copies /etc/hosts from host during starting container,
we would like to keep /etc/hosts file updated since it is managed by salt.
The reason for this is that velum has to resolve Kube API fqdn.
Problem was discovered and discribed in bug 1062728

### DIFF
--- a/public.yaml
+++ b/public.yaml
@@ -308,6 +308,9 @@ spec:
       - mountPath: /var/lib/misc/infra-secrets
         name: infra-secrets
         readOnly: True
+      - mountPath: /etc/hosts
+        name: etc-hosts
+        readOnly: True
     args: ["bin/init"]
   - name: openldap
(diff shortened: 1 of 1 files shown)


inercia:
@mjura are you sure this works? `/etc/hosts` is managed by Docker, so I guess you cannot just mount it...

mjura:
yes, it works, I'm still testing it

mjura:
Fix will be done in another way, moving to
https://github.com/kubic-project/salt/pull/265

flavio:
This should work just fine, on my host docker handles that in the expected way:

```
docker run -ti --rm -v /etc/hosts:/etc/hosts:ro opensuse /bin/bash
```

That should work also with kubernetes managing the container.

PR #132, opened 2017-10-24
Create velum_test database

flavio:
Ensure the mariadb init container creates also the `velum_test` database, this is required to have devenv run the test suite against a database which is **not** the production one.

Required by https://github.com/kubic-project/automation/pull/180

### DIFF
--- a/setup/mysql/setup-mysql.sh
+++ b/setup/mysql/setup-mysql.sh
@@ -25,11 +25,15 @@ done
 velum_passwd=`cat /infra-secrets/mariadb-velum-password`
 salt_passwd=`cat /infra-secrets/mariadb-salt-password`
 
+# Create both the ENV and the test databases.
+# This is requird to run the test suites inside of devenv
 mysql $mysql_flags -f <<EOF
   CREATE SCHEMA IF NOT EXISTS velum_$ENV;
+  CREATE SCHEMA IF NOT EXISTS velum_test;
   CREATE USER velum@localhost IDENTIFIED BY "$velum_passwd";
   CREATE USER salt@localhost IDENTIFIED BY "$salt_passwd";
   GRANT ALL PRIVILEGES ON velum_$ENV.* TO velum@localhost;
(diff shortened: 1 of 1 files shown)


ereslibre:
Wow, I'm not sure we want the `velum_test` database to be present in production environments... I would try to try a different take on this problem but not sure if it will be more "hackish" or *when* I will be able to give it a try, I don't want to block others in the meantime though...

ereslibre:
I was thinking in adding something in `caasp-kvm` or `caasp-devenv` (the automation one of course) that executed `setup/mysql/setup-mysql.sh` with `test` `ENV`, so it would automatically generate this database (without having to modify this bit here).

setup/mysql/setup-mysql.sh
```diff
@@ -25,11 +25,15 @@ done
 velum_passwd=`cat /infra-secrets/mariadb-velum-password`
 salt_passwd=`cat /infra-secrets/mariadb-salt-password`
 
+# Create both the ENV and the test databases.
+# This is requird to run the test suites inside of devenv
```
MaximilianMeister: small typo `requird`

MaximilianMeister:
> I was thinking in adding something in caasp-kvm or caasp-devenv (the automation one of course) that executed setup/mysql/setup-mysql.sh with test ENV, so it would automatically generate this database (without having to modify this bit here).

if it's possible i'd also do that, as we don't ship any unittests anyways, so the db would be rather useless in a production env

flavio:
Yeah, I'm not convinced as well to have `velum_test` on production environments too. I tried to run the `setup/mysql/setup-mysql.sh` with the `ENV=test` but this is not going to work because the `velum` and `salt` users already exists.

Our version of mariadb doesn't support the `CREATE USER IF NOT EXISTS` statement, which makes things more complicated.

I'll create a new script, just for the `test` environment to be invoked by the devenv

flavio:
@ereslibre oh, I'll give it a try

flavio:
The recommendation given by @ereslibre works, there's no need for this PR anymore. Closing.

PR #134, opened 2017-11-23
Use HAProxy from the beginning for Velum too.

ereslibre:
This will help us with SSL termination on HAProxy side.

- [x] Do not call to `bin/init` from two different containers, as they will setup the database the first time and run migrations afterwards. It's safer if only one container runs this while the others just raise the application server.

Depends on: https://github.com/kubic-project/velum/pull/387
Depends on: https://github.com/kubic-project/salt/pull/314

### DIFF
--- a/admin-node-setup.sh
+++ b/admin-node-setup.sh
@@ -32,11 +32,16 @@ if [ ! -f $manifest_dir/private.yaml ]; then
     echo "private.yaml is not in $manifest_dir" >&2
     exit -3
 fi
+if [ ! -f $manifest_dir/haproxy.yaml ]; then
+    echo "haproxy.yaml is not in $manifest_dir" >&2
+    exit -3
+fi
 
 tmp_dir=$(mktemp -d)
 
 cp $manifest_dir/public.yaml $tmp_dir
 cp $manifest_dir/private.yaml $tmp_dir
+cp $manifest_dir/haproxy.yaml $tmp_dir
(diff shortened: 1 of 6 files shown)


ereslibre:
How to test: https://github.com/kubic-project/salt/pull/314#issuecomment-346907208

flavio:
I didn't test the change, but it looks good to me. I've only one question: shouldn't we move all the velum pods definitions from `public.yml` to either `private.yml` or to a new `.yml` file? Originally `public.yml` has all the pods that are reusing the host network, but this is no longer needed now by velum; hence I would like to play safe and avoid to reuse the host network stack with it.

ereslibre:
> I didn't test the change, but it looks good to me. I've only one question: shouldn't we move all the velum pods definitions from public.yml to either private.yml or to a new .yml file? Originally public.yml has all the pods that are reusing the host network, but this is no longer needed now by velum; hence I would like to play safe and avoid to reuse the host network stack with it.

I agree, I think we should explore this, but I'd do this in a separate PR.

kiall:
@flavio - yep - I've been thinking the same and started with this PR - https://github.com/kubic-project/caasp-container-manifests/pull/135 - that gets the package into a place where new yaml files will work in our dev env, without needing a new c-c-m package built :)

kiall:
Force merging the 3 PRs in together.

PR #135, opened 2017-11-24
Move manifests into a subdirectory

kiall:
This allows us to remove the public.yaml / private.yaml hardcodes, which in
turn allows us to split the public / private pods into more specific pods
matching the typical patterns used to deploy workloads in K8S.

Depends-On: https://github.com/kubic-project/automation/pull/191


ereslibre:
This PR has now some conflicts.

PR #138, opened 2017-12-14
Do not use FileOrCreate resource type.

ereslibre:
The init container for the secrets will create this file.

### DIFF
--- a/manifests/public.yaml
+++ b/manifests/public.yaml
@@ -89,6 +89,7 @@ spec:
   - name: salt-master-config
     image: sles12/velum:__TAG__
     command: ["sh", "-c", "umask 377;
+                           rmdir /salt-master-credentials/55-returner-credentials.conf;
                            if ! grep mysql /salt-master-credentials/55-returner-credentials.conf; then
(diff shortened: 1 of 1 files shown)


manifests/public.yaml
```diff
@@ -89,6 +89,7 @@ spec:
   - name: salt-master-config
     image: sles12/velum:__TAG__
     command: ["sh", "-c", "umask 377;
+                           rmdir /salt-master-credentials/55-returner-credentials.conf;
```
mjura: Why this is directory and not file?

ereslibre: The very first time k8s loads this manifest, this file does not exist on the host path, and is created as a directory. The init container removes this directory and writes the file. It's far from ideal, but does the trick.

ereslibre:
This isn't passing on CI because the orchestration never finishes (but Velum is started with this patch). Now the problem is related to Dex and the update of kubernetes from 1.7.7 to 1.8.5.

manifests/public.yaml
inercia: wtf! I think you should add a comment explaining this...

PR #140, opened 2018-01-17
Disable swap

FedericoCeratto:
Related to https://github.com/kubic-project/salt/pull/345

 bsc#1075001

### DIFF
--- a/activate.sh
+++ b/activate.sh
@@ -28,3 +28,7 @@ roles:
 EOF
 echo "master: localhost" > /etc/salt/minion.d/minion.conf
 echo "id: admin" > /etc/salt/minion.d/minion_id.conf
+
+# Disable swap
+# On ISO-based installs the script runs in a chroot and then the system is rebooted
+sed -i '/^#/! {/ swap / s/^/#/}' /etc/fstab
--- a/admin-node-setup.sh
+++ b/admin-node-setup.sh
@@ -6,6 +6,9 @@ if [ -f /etc/salt/grains ]; then
     sed -i -e 's|tx_update_reboot_needed:.*|tx_update_reboot_needed: false|g' /etc/salt/grains
 fi
 
+# Turn off current swaps if any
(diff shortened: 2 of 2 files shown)


activate.sh
```diff
@@ -28,3 +28,8 @@ roles:
 EOF
 echo "master: localhost" > /etc/salt/minion.d/minion.conf
 echo "id: admin" > /etc/salt/minion.d/minion_id.conf
+
+# Disable swap
+# On ISO-based installs the script runs in a chroot and then the system is rebooted
+sed -i 's/^\(.* swap [ ]* sw .*\)$/#\1/' /etc/fstab
+sed -i 's/^\(.* swap [ ]* swap .*\)$/#\1/' /etc/fstab
```
mjura: We could merge this two lines into
```
sed -i 's/^\(.* swap .*\)$/#\1/' /etc/fstab
```

mjura: Or
```
sed -i '/ swap / s/^/#/' /etc/fstab
```

FedericoCeratto: I wanted to explicitly match ' swap ' followed by ' swap ' or ' sw ' (with whitespaces around them) in a more readable way.

mjura: The partition type is always  swap ` with whitespaces, so we can search only for this string. Other options can be set in different way so it is not worth to worry about it.

FedericoCeratto: " swap " might potentially appear in a comment - that's why I was matching for the second sw/swap

mjura: This will skip already commented lines with `swap` string inside
```
sed -i ' /^#/! / swap /s/^/#/' /etc/fstab
```

Problem which I see here is that these regexp are not covering following examples:
```
/dev/sda2 none swap sw,pri=1 0 0
```
or
```
/dev/sda3 none swap defaults 0 0
```

mjura:
We need backport it to `release-2.0`

PR #142, opened 2018-01-24
Add mount for public cloud pillar

jgleissner:
To get public cloud specific configuration data to be used with salt-cloud into the salt-master container, we'd like to map the directory where we store the generated sls file into the container.


jgleissner:
See also https://github.com/kubic-project/salt/pull/358

PR #144, opened 2018-01-31
Add kubic images

jordimassaguerpla:
If this package is built in tumbleweed,the images should be named
tumbleweed and not sle12.

When this builds on sle15, the images should be named sle15.

We cannot use the same name for the different suse versions as if this
images were in a registry, they should be named different.

### DIFF
--- a/packaging/suse/make_spec.sh
+++ b/packaging/suse/make_spec.sh
@@ -37,6 +37,18 @@ cat <<EOF > ${NAME}.spec
 # Please submit bugfixes or comments via http://bugs.opensuse.org/
 #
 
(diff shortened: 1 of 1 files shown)


kiall:
In addition to this, there are other places where we reference sles12/* e.g.

https://github.com/kubic-project/salt/blob/master/salt/dex/dex.yaml#L102
https://github.com/kubic-project/salt/blob/master/salt/addons/dns/kubedns.yaml.jinja#L34
https://github.com/kubic-project/salt/blob/master/salt/addons/tiller/tiller.yaml.jinja#L29

PR #148, opened 2018-02-06
Use base_image name as 'kubic-' for kubic images

sysrich:
After discussion, we collectively decided kubic- makes more sense as the _base_image name for kubic images, even if they're built from Tumbleweed


jordimassaguerpla:
@sysrich : You need to do the same in the salt project

PR #149, opened 2018-02-08
Sync haproxy manifest with salt repo

kiall:
The haproxy manifest is duplicated between the salt and c-c-m repo,
sync the recent changes from the salt repo over here to keep
everything lined up.

Syncs 25c660fd92150fbf8b1a7213282d2f9ead9a67e6 from the salt repo.

### DIFF
--- a/manifests/haproxy.yaml
+++ b/manifests/haproxy.yaml
@@ -6,6 +6,8 @@ metadata:
   namespace: kube-system
   labels:
     name: haproxy
+  annotations:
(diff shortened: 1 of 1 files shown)


kiall:
There is no need to backport this one to 2.0 - as the 2.0's c-c-m doesn't include this manifest.

flavio:
Ouch, I wasn't aware of this duplication. We should get rid of it. In the meantime I'll merge the PR

PR #152, opened 2018-02-14
Fix three upgrade issues

kiall:
* Migrate the old HAProxy config over to the new path
* Add the new static velum/velum-api haproxy sections
* Generate the missing *-bundle.pem files

Fixes bsc#1080978

### DIFF
--- a/admin-node-setup.sh
+++ b/admin-node-setup.sh
@@ -49,13 +49,62 @@ cp $tmp_dir/*.yaml $kube_dir/
 
 rm -rf $tmp_dir
 
-# First time setup of user-configuration for salt-master
+# Create CaaSP config dir
 if [ ! -d "/etc/caasp" ]; then
     mkdir /etc/caasp
 fi
 
+# First time setup of user-configuration for salt-master
 if [ ! -f "/etc/caasp/salt-master-custom.conf" ]; then
     echo "# Custom Configurations for Salt-Master" > /etc/caasp/salt-master-custom.conf
+fi
+
+# Migrate haproxy config path post path change
+if [[ ! -f "/etc/caasp/haproxy/haproxy.cfg" && -f "/etc/haproxy/haproxy.cfg" ]]; then
+    if [ ! -d "/etc/caasp/haproxy" ]; then
+        mkdir /etc/caasp/haproxy
+    fi
+
+    mv /etc/haproxy/haproxy.cfg /etc/caasp/haproxy/haproxy.cfg
+
+    # Add the Velum and Velum-API services to HAproxy
+    cat << EOF >> /etc/caasp/haproxy/haproxy.cfg
+
+listen velum
+        bind 0.0.0.0:80
+        bind 0.0.0.0:443 ssl crt /etc/pki/velum.pem ca-file /etc/pki/ca.crt
+        mode http
+        acl path_autoyast path_reg ^/autoyast$
+        option forwardfor
+        http-request set-header X-Forwarded-Proto https
+        redirect scheme https code 302 if !{ ssl_fc } !path_autoyast
+        default-server inter 10s fall 3
+        balance roundrobin
+        server velum unix@/var/run/puma/dashboard.sock
+
+listen velum-api
+        bind 127.0.0.1:443 ssl crt /etc/pki/velum.pem ca-file /etc/pki/ca.crt
+        mode http
+        option forwardfor
+        http-request set-header X-Forwarded-Proto https
+        default-server inter 10s fall 3
+        balance roundrobin
+        server velum unix@/var/run/puma/api.sock
+EOF
+fi
+
+# Generate missing TLS bundle files
+if [ ! -f "/etc/pki/private/velum-bundle.pem" ]; then
+    cat /etc/pki/velum.crt /etc/pki/private/velum.key > /etc/pki/private/velum-bundle.pem
+    chmod 600 /etc/pki/private/velum-bundle.pem
+fi
+if [ ! -f "/etc/pki/private/salt-api-bundle.pem" ]; then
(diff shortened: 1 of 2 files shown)


admin-node-setup.sh
```diff
@@ -49,15 +49,61 @@ cp $tmp_dir/*.yaml $kube_dir/
...
+        http-request set-header X-Forwarded-Proto https
+        default-server inter 10s fall 3
+        balance roundrobin
+        server velum unix@/var/run/puma/api.sock
+EOF
+fi
+
+# Generate missing TLS bundle files
+if [ ! -f "/etc/pki/private/velum-bundle.pem" ]; then
+    cat /etc/pki/velum.crt /etc/pki/private/velum.key > /etc/pki/private/velum-bundle.pem
```
mjura: We should set mask `600` on this file

admin-node-setup.sh
```diff
@@ -49,15 +49,61 @@ cp $tmp_dir/*.yaml $kube_dir/
...
+fi
+if [ ! -f "/etc/pki/private/salt-api-bundle.pem" ]; then
+    cat /etc/pki/salt-api.crt /etc/pki/private/salt-api.key > /etc/pki/private/salt-api-bundle.pem
```
mjura: We should set mask 600 on this file

admin-node-setup.sh
```diff
@@ -49,15 +49,61 @@ cp $tmp_dir/*.yaml $kube_dir/
...
+fi
+if [ ! -f "/etc/pki/private/ldap-bundle.pem" ]; then
+    cat /etc/pki/ldap.crt /etc/pki/private/ldap.key > /etc/pki/private/ldap-bundle.pem
```
mjura: We should set mask 600 on this file

kiall:
> does it make sense to put upgrade related code into a separate script/functions just for organizational reasons? i suspect this could grow over time and versions, and ideally we could be more specific about which code is needed for what upgrade. other than that LGTM

I suspect the right thing to do is use salt in it's masterless mode to configure the admin node, in place of these awful scripts :)

kiall:
[Bug: 1080978  -  v2 to v3 upgrade fails](https://trello.com/c/1eXcvN27/633-bug-1080978-v2-to-v3-upgrade-fails)

admin-node-setup.sh
```diff
@@ -49,15 +49,61 @@ cp $tmp_dir/*.yaml $kube_dir/
...
+
+listen velum-api
+        bind 127.0.0.1:443 ssl crt /etc/pki/velum.pem ca-file /etc/pki/ca.crt
+        mode http
+        option forwardfor
+        http-request set-header X-Forwarded-Proto https
+        default-server inter 10s fall 3
+        balance roundrobin
+        server velum unix@/var/run/puma/api.sock
+EOF
```
ereslibre: Why is this append required? If the package installed the new configuration, `/etc/caasp/haproxy/haproxy.cfg` should include https://github.com/kubic-project/caasp-container-manifests/blob/master/config/haproxy/haproxy.cfg with `velum` and `velum-api` already, no? In 2.0 this file does not exist, but during the package upgrade, `/etc/caasp/haproxy/haproxy.cfg` should have been created, no?

admin-node-setup.sh
```diff
@@ -49,15 +49,61 @@ cp $tmp_dir/*.yaml $kube_dir/
 
 rm -rf $tmp_dir
 
-# First time setup of user-configuration for salt-master
+# Create CaaSP config dir
 if [ ! -d "/etc/caasp" ]; then
     mkdir /etc/caasp
 fi
```
ereslibre: Shouldn't this be created by the package?

admin-node-setup.sh
```diff
@@ -49,15 +49,61 @@ cp $tmp_dir/*.yaml $kube_dir/
...
+listen velum-api
+        bind 127.0.0.1:443 ssl crt /etc/pki/velum.pem ca-file /etc/pki/ca.crt
+        mode http
+        option forwardfor
+        http-request set-header X-Forwarded-Proto https
+        default-server inter 10s fall 3
+        balance roundrobin
+        server velum unix@/var/run/puma/api.sock
+EOF
+fi
```
ereslibre: Maybe I'm misunderstanding, but do we really need to move the configuration from the old location to the new? When the package with the new location is installed, shouldn't this config be installed as part of the package installation? 

https://github.com/kubic-project/caasp-container-manifests/commit/e8ace8fe824383bc0118520cf20de07d0358d564 and https://github.com/kubic-project/caasp-container-manifests/commit/d707d7dda42d18cb1e4587a9b63a61a6796c2250

kiall: Yes, we need to move it. If we revert to the default configuration,  then we end up removing the Dex and Kube-API proxies. These will come back once the user eventually upgrades all nodes, but, we should keep it working for the period where they haven't done that.

admin-node-setup.sh
kiall: The packaged file only shows up in a fresh install, during an upgrade, tx update has no access to /etc (well, it does, just those changes will disappear into the ether once the tx-update has exited)

admin-node-setup.sh
kiall: Not on upgrade, tx update can't make any persisted changes to /etc

ereslibre: I didn't realize about this until now. This will be tricky if we include a new package in the future that comes with its own `/etc` file and that file(s) are not managed by salt explicitly (because we don't want to tune them).

ereslibre: Or if a system package updates its `/etc` config for let's say more secure defaults... guh.

ereslibre: Okay, it's an overlay mount.

admin-node-setup.sh
kiall: Fixed all 3, + the gen-certs.sh script.

PR #153, opened 2018-02-14
enable certificate validation for net-ldap

MaximilianMeister:
CVE-2017-17718 requires net-ldap to validate the certificate

therefore set a fixed resolvable name for ldap and generate
the certificate for it

### DIFF
--- a/admin-node-setup.sh
+++ b/admin-node-setup.sh
@@ -61,6 +61,12 @@ fi
 # Generate TLS CA and Initial Certificates
 /usr/share/caasp-container-manifests/gen-certs.sh
 
+# add an entry for ldap.infra.caasp.local to /etc/hosts
+# this is needed to enable net-ldap to validate the certificate for LDAP_HOST
+if ! [ "$(cat /etc/hosts | grep -E "^127.0.0.1\s+" | grep ldap.infra.caasp.local)" ]; then
+    sed -i 's/127.0.0.1\tlocalhost/127.0.0.1\tlocalhost ldap.infra.caasp.local/g' /etc/hosts
+fi
+
 VELUM_CRT_FINGERPRINT_SHA1=$(openssl x509 -noout -in /etc/pki/velum.crt -fingerprint -sha1 | cut -d= -f2)
 VELUM_CRT_FINGERPRINT_SHA256=$(openssl x509 -noout -in /etc/pki/velum.crt -fingerprint -sha256 | cut -d= -f2)
 
--- a/gen-certs.sh
+++ b/gen-certs.sh
@@ -161,4 +161,4 @@ set -e
 genca
 gencert "velum" "Velum" "$all_hostnames" "$(ip_addresses)"
 gencert "salt-api" "salt-api.infra.caasp.local" "" "127.0.0.1"
-gencert "ldap" "OpenLDAP" "$all_hostnames" "$(ip_addresses)"
+gencert "ldap" "OpenLDAP" "ldap.infra.caasp.local" "$(ip_addresses)"
--- a/manifests/public.yaml
+++ b/manifests/public.yaml
@@ -256,7 +256,7 @@ spec:
     - name: VELUM_INTERNAL_API_PASSWORD_FILE
       value: /var/lib/misc/infra-secrets/velum-internal-api-password
     - name: LDAP_HOST
-      value: "127.0.0.1"
+      value: "ldap.infra.caasp.local"
     - name: LDAP_PORT
       value: "389"
     - name: LDAP_GROUP_BASE_DN
@@ -326,7 +326,7 @@ spec:
     - name: VELUM_INTERNAL_API_PASSWORD_FILE
       value: /var/lib/misc/infra-secrets/velum-internal-api-password
     - name: LDAP_HOST
-      value: "127.0.0.1"
+      value: "ldap.infra.caasp.local"
     - name: LDAP_PORT
       value: "389"
     - name: LDAP_GROUP_BASE_DN


MaximilianMeister:
needed for https://github.com/kubic-project/velum/pull/443

admin-node-setup.sh
```diff
@@ -61,6 +61,10 @@ fi
 # Generate TLS CA and Initial Certificates
 /usr/share/caasp-container-manifests/gen-certs.sh
 
+# add an entry for ldap.infra.caasp.local to /etc/hosts
+# this is needed to enable net-ldap to validate the certificate for LDAP_HOST
+sed -i 's/127.0.0.1\tlocalhost/127.0.0.1\tlocalhost ldap.infra.caasp.local/g' /etc/hosts
```
kiall: We'll need to make this a little more robust - if ran over and over (as `admin-node-setup.sh` is on each reboot), we'll end up adding this new entry multiple times.

kiall: We'll also want to update salt's salt/etc-hosts/hosts.jinja to make sure that doesn't remove this entry from the admin node.

gen-certs.sh
```diff
@@ -161,4 +161,4 @@ set -e
 genca
 gencert "velum" "Velum" "$all_hostnames" "$(ip_addresses)"
 gencert "salt-api" "salt-api.infra.caasp.local" "" "127.0.0.1"
-gencert "ldap" "OpenLDAP" "$all_hostnames" "$(ip_addresses)"
+gencert "ldap" "ldap.infra.caasp.local" "" "127.0.0.1"
```
kiall: The "127.0.0.1" here is making the cert valid for "127.0.0.1" - which is generally considered bad practice at best, and a security issue at worst :)

admin-node-setup.sh
kiall: Something like this maybe?

	if [ "$(cat /etc/hosts | grep -E "^127.0.0.1\s+" | grep ldap.infra.caasp.local)" ]; then
	    echo "Have it already"
	else
	    echo "Don't have it"
	fi

gen-certs.sh
MaximilianMeister: oops, that was still remainder of testing

admin-node-setup.sh
MaximilianMeister: will we still need this after bootstrap? i am not fully sure, but for now i think it's only needed when creating the ldap user, but i might be wrong

MaximilianMeister:
[(2) move factory's devel project to d:l:r:e for velum dependent gems](https://trello.com/c/adVQVCXk/569-2-move-factorys-devel-project-to-dlre-for-velum-dependent-gems)

MaximilianMeister:
[(3) [security] update net-ldap because of CVE-2017-17718](https://trello.com/c/2jzG8pSr/511-3-security-update-net-ldap-because-of-cve-2017-17718)

MaximilianMeister:
rebased

manifests/public.yaml
```diff
@@ -326,7 +326,7 @@ spec:
     - name: VELUM_INTERNAL_API_PASSWORD_FILE
       value: /var/lib/misc/infra-secrets/velum-internal-api-password
     - name: LDAP_HOST
-      value: "127.0.0.1"
+      value: "ldap.infra.caasp.local"
```
ereslibre: So, during the highstates the ldap certificate gets generated by salt:

https://github.com/kubic-project/salt/blob/f29c60a90ecccf49089255267a119ede4c1f6032/salt/ldap/init.sls#L7-L12

I see that after this certificate gets recreated, from a running cluster:

```
            X509v3 Subject Alternative Name: 
                DNS:caasp-admin, DNS:caasp-admin.infra.caasp.local, DNS:be19dde4e72240bdbf9939d4720edec5, DNS:be19dde4e72240bdbf9939d4720edec5.infra.caasp.local, IP Address:10.17.1.0
```

I think we should also add `ldap.infra.caasp.local` to `ldap/init.sls` so Velum will correctly validate the certificate after the bootstrap has finished, since with the current situation I think Velum should fail to connect to LDAP after the bootstrap has finished and after we have regenerated the ldap certificate.

MaximilianMeister: https://github.com/kubic-project/salt/pull/396

admin-node-setup.sh
```diff
@@ -61,6 +61,12 @@ fi
 # Generate TLS CA and Initial Certificates
 /usr/share/caasp-container-manifests/gen-certs.sh
 
+# add an entry for ldap.infra.caasp.local to /etc/hosts
+# this is needed to enable net-ldap to validate the certificate for LDAP_HOST
+if ! [ "$(cat /etc/hosts | grep -E "^127.0.0.1\s+" | grep ldap.infra.caasp.local)" ]; then
+    sed -i 's/127.0.0.1\tlocalhost/127.0.0.1\tlocalhost ldap.infra.caasp.local/g' /etc/hosts
```
MaximilianMeister: @ereslibre do we also need to adapt `/etc/hosts` in salt? i am not fully sure if it is still relevant *after* creating the ldap user, but the file will likely be overwritten by salt

admin-node-setup.sh
ereslibre: Yes, the Velum login for now connects to the only internal LDAP using the `LDAP_HOST` envvar, so the `/etc/hosts` needs to resolve whatever that is.

admin-node-setup.sh
ereslibre: Yes, it's relevant because you want to login to Velum afterwards. Velum needs to reach the LDAP instance. Before it was `127.0.0.1`, but with this change, `ldap.infra.caasp.local` needs to be "resolvable" at all times, otherwise the login from Velum will fail.

ereslibre: Note that I'm talking about the very first login page (not the kubeconfig download that uses dex and that's fine).

MaximilianMeister: added in https://github.com/kubic-project/salt/pull/396

kiall:
This looks good now - however, it should not merge until after the salt change has merged - will update my review once that is done!

PR #156, opened 2018-02-16
[WIP] Remove VELUM_WORKERS environment variable

kiall:
Velum had been updated to run more workers, but that didn't take
effect as this was overriding the default.

The default will now be applied - which is 8 workers for both the
dashboard, and api containers.

----

The result of this will be more memory used... are we OK with that?

### DIFF
--- a/manifests/public.yaml
+++ b/manifests/public.yaml
@@ -233,8 +233,6 @@ spec:
       value: /var/lib/misc/velum-secrets
     - name: VELUM_SOCKET_NAME
       value: dashboard.sock
-    - name: VELUM_WORKERS
-      value: "2"
     - name: VELUM_DB_HOST
       value:
     - name: VELUM_DB_USERNAME
@@ -303,8 +301,6 @@ spec:
       value: /var/lib/misc/velum-secrets
     - name: VELUM_SOCKET_NAME
       value: api.sock
-    - name: VELUM_WORKERS
-      value: "6"
     - name: VELUM_DB_HOST
       value:
     - name: VELUM_DB_USERNAME


flavio:
I agree with @MaximilianMeister's first point. I'm not sure it's a good idea to take away this feature, it makes possible to be pretty flexible...

kiall:
> I'd stick with the environment variable and set it to whatever is the default. that's more explicit and at a customer's instance you could more easily tweak the setting as it's already there

Doing this is entirely unsupported as we'll go ahead and revert the manual change on the next admin node reboot - if anything it gives a false sense of a feature that doesn't exist!

> i was wondering as we are using k8s to "serve" velum if we can't make the number of workers/threads dynamic. generally in puma you set the nr of workers to whatever the number of cpu's (respecting memory as well) the host machine has, and the number of threads to what you expect the application to handle in terms of concurrent requests.

Dynamically setting the workers/threads based on CPU / RAM counts etc is typically only done when a machine is dedicated to that software - given we have a whole bunch of different things on the machine, we don't want to over-consume resources just for Velum (which this PR could very well be doing by starting more workers than was happening previously)

ereslibre:
The only reason for this envvar to exist was because the `api` and the `dashboard` would have different needs in terms of peak requests. However they haven't been properly measured since the beginning, so I'm not against defaulting both back to whatever puma says (as @kiall said, we increased the workers in puma but they didn't really have any effect as that setting was overriden here).

MaximilianMeister:
> Doing this is entirely unsupported as we'll go ahead and revert the manual change on the next admin node reboot - if anything it gives a false sense of a feature that doesn't exist!

not really, it's not exposed in any way as a feature to the customer, it just would be easier for the supporter to tweak this value. and we likely might run into that in certain environments. e.g. if you have 8 workers on a system with only 4 cores, which (on top of being virtual/containerized) might just be a wrong configuration.

i think 8 workers is way too much, is there any reason for this high value?

kiall:
> not really, it's not exposed in any way as a feature to the customer, it just would be easier for the supporter to tweak this value.

If support tweaks the value, we will undo that tweak - it's not something that is tunable.

> i think 8 workers is way too much, is there any reason for this high value?

Yea, this is an open question... is 8 too much? If it is - we should IMO merge this, and update https://github.com/kubic-project/velum/blob/master/config/puma.rb#L1 to the value we actually want.

MaximilianMeister:
> If support tweaks the value, we will undo that tweak - it's not something that is tunable.

i think that's clear. nevertheless it would be helpful IMO during debugging a machine when there's a possibly related issue, e.g. memory/cpu over-consumption or puma choking on a flood of requests in environments where there are many nodes, to try if a different number of workers fixes the issue... just making things up here but from past experience, i think nr of workers are a crucial part of configuration when it comes to highly dynamic environments and resulting different loads of requests to the webserver.

other than that what's the point of doing `WORKERS = ENV.fetch("VELUM_WORKERS", 8)` and not hardcoding it all the way via `WORKERS = 8` - at least until now every similar project i've seen so far uses the ENV || default way because of that reasons, so i'd keep it and use a saner (lower) default if there's no other reason, i couldnt find any reason in the commit message for the 2 -> 8 increase in https://github.com/kubic-project/velum/commit/c5637d2d8388b1d96b28f786eb798a1d523f2e88

ereslibre:
> i couldnt find any reason in the commit message for the 2 -> 8 increase in kubic-project/velum@c5637d2

That was my bad. I cannot fully remember but I think part of this effort was when we tried to use the pillar api without haproxy and rails serving the ssl/tls requests on its own, which was pretty bad.

We never did a clear insight of how many workers and how many threads were required and for what usage (browser, api requests [depending on cluster size])... It has always been magic numbers that we have put there.

MaximilianMeister:
> We never did a clear insight of how many workers and how many threads were required and for what usage (browser, api requests [depending on cluster size])... It has always been magic numbers that we have put there.

yeah, we should probably do that at some point. we only had one bug (bsc#1048533) where we needed to increase the workers initially from `0` to `2`. `0` is the [default](https://github.com/puma/puma/blob/master/examples/config.rb#L110) so i'd try to go back to `2` and see if we run into issues again, what do you think?

flavio:
> other than that what's the point of doing WORKERS = ENV.fetch("VELUM_WORKERS", 8) and not hardcoding it all the way via WORKERS = 8 - at least until now every similar project i've seen so far uses the ENV || default way because of that reasons, so i'd keep it and use a saner (lower) default if there's no other reason, i couldnt find any reason in the commit message for the 2 -> 8 increase in kubic-project/velum@c5637d2

+1

ereslibre:

PR #157, opened 2018-02-20
Make entrypoint of mariadb-user-secrets container more robust

flavio:
I've run into a timing issue that caused the root password of mariadb **not** being injected into the running container "mariadb-user-secrets" in time. That caused the container to enter an infinite loop consisting of trying to connect to mariadb as root without a specifying password, getting an error message, sleeping 1 second and trying again.

This is an init container, as long as it's running kubelet won't start over containers, like openldap, velum-*, salt-*,...

With this change the root password file constantly read before attempting a new connection.


davidcassany:
@flavio, not sure I understand the issue, but is it related to the time that the mariadb container takes to be up and running with a defined root password? Just asking because probably there is room for improvement in the mariadb entrypoint script or container definition in order to get a faster _boot_.

flavio:
The database was up and running and the root account was already protected by a password.
For some unknown reason the `mariadb-user-secrets` container was trying to access the database using a password-less root user. That resulted in a failure which lead to an endless loop.

That happened because, for some unknown reasons, the `mariadb-user-secrets` didn't get the root password via the `/infra-secrets/mariadb-root-password` file.
I logged into the running container (via `docker exec`) and I found a regular /infra-secrets/mariadb-root-password` (it had the root password).

I logged out of the container and I killed it via `docker kill`. That caused the kubelet to restart the container immediately, this time it succeeded and, as a result of that, the whole velum stack came up.

I suspect this might be a corner case stressed out by the performance troubles we are facing because spectre and meltdown.

@ereslibre: I see what you mean. I'll change the code accordingly.

flavio:
@ereslibre code changed, please take a look again

ereslibre:
@flavio isn't it the same?

flavio:
I didn't notice this line. Shouldn't we change that to check the file has actually some contents? Maybe (due to spectre/meltdown) the code that generates the random password took more time than expected...

ereslibre:
> I didn't notice this line. Shouldn't we change that to check the file has actually some contents? Maybe (due to spectre/meltdown) the code that generates the random password took more time than expected...

So is there a chance in which the file was created but the contents were not yet synced maybe? It's the only thing I can think about.

flavio:
@ereslibre I changed the code to wait until the file exists and is not empty. This shouldn't make the code look so weird

setup/mysql/setup-mysql.sh
```diff
@@ -1,7 +1,8 @@
 #!/usr/bin/env bash
 umask 377;
 
-while [ ! -f /infra-secrets/mariadb-root-password ]; do
+# wait untile the file exists and has contents
```
ereslibre: Oops, small typo `until`.

flavio:
I fixed the typo

flavio:
The failure of Jenkins is not related with this change. Merging.

PR #160, opened 2018-02-27
Haproxy: Mount config dir as read only

kiall:
There is no good reason for this container to be modifying anything in
/etc/haproxy (or /etc/caasp/haproxy as seen on the host). Remove it's
ability to do so.


kiall:
This will probably fail CI as the container entrypoint is currently doing this:

    sed -Ei 's/^\s*daemon$//' $CONFIG

PR #161, opened 2018-03-07
Change manifest __TAG__'s for Kubic also

sysrich:
Tested on latest Kubic image, proven to work (now the Admin node is broken for wholly more interesting reasons)


sysrich:
Is there a reason this still hasn't been merged?

PR #162, opened 2018-03-15
Remove Kubic workaround, caasp-tools no longer conflicts

sysrich:
Urgent - there is a mismatch between what is in our git repo and what is in OBS - this resolves that.

Any caasp-container-manifests submissions before this gets submitted to Factory risk breaking Kubic


jordimassaguerpla:
I was waiting for the tests to finish but they fail. Still I don't think they are related, since your change is related to the package.

PR #163, opened 2018-03-16
Rename 90-velum.conf to 80-velum.conf to avoid conflict on Kubic

thkukuk:
and SLE15 and get the same output as on CaaSP.


ereslibre:
Since `admin-node-setup.sh` is executed on every boot, isn't this going to produce now another issue file with the same contents if it's an update? I think for that case we should check for its existence and just `mv` it.

thkukuk:
If this file is created during every boot, the file should not be stored in /etc/issue.d, but the correct directory would be /run/issue.d/

thkukuk:
Funny to request things in the commit message which are standing in the commit message ...

PR #171, opened 2018-04-16
salt-master and salt-api should not load each others configuration files

kiall:
Since both need different log_level settings, they need their own config
files. The intent for these two files was just that - api specific, and
master specific settings.

Related to https://github.com/kubic-project/salt/pull/486

### DIFF
--- a/manifests/public.yaml
+++ b/manifests/public.yaml
@@ -134,9 +134,6 @@ spec:
     - mountPath: /etc/salt/master.d/50-master.conf
       name: salt-master-config-master-conf
       readOnly: True
-    - mountPath: /etc/salt/master.d/50-api.conf
-      name: salt-master-config-api-conf
-      readOnly: True
     - mountPath: /etc/salt/master.d/50-peer.conf
       name: salt-master-config-peer-conf
       readOnly: True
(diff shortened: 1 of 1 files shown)


kiall:
I wonder if anything in 50-master.conf is needed by `salt-api`? I'm not sure our separation remained as clean as we wanted it to!

ereslibre:
I think the api needs `50-master.conf` but I'm not completely sure about it at the moment.

MaximilianMeister:
> I think the api needs 50-master.conf but I'm not completely sure about it at the moment.

this is more likely than the case where the master.conf needs the api.conf.

i can make a testrun, but i think only removing the api.conf from the master already is sufficient, as we wont have info log_level on master anymore

manifests/public.yaml
```diff
@@ -181,9 +178,6 @@ spec:
     - mountPath: /etc/pki/ca.crt
       name: ca-certificate
       readOnly: True
-    - mountPath: /etc/salt/master.d/50-master.conf
-      name: salt-master-config-master-conf
-      readOnly: True
```
MaximilianMeister: > i can make a testrun,

ok, `/setup/discovery` immediately throws `Velum::SaltApi::SaltConnectionException` I guess it's because of this change

manifests/public.yaml
```diff
@@ -134,9 +134,6 @@ spec:
     - mountPath: /etc/salt/master.d/50-master.conf
       name: salt-master-config-master-conf
       readOnly: True
-    - mountPath: /etc/salt/master.d/50-api.conf
-      name: salt-master-config-api-conf
-      readOnly: True
```
MaximilianMeister: i'll run another test with only this change

MaximilianMeister: salt-api

MaximilianMeister: we could just move this block https://github.com/kubic-project/salt/blob/master/config/master.d/50-api.conf#L4 to `50-master.conf` and only keep the `log_level` in `50-api.conf`

kiall: > we could just move this block https://github.com/kubic-project/salt/blob/master/config/master.d/50-api.conf#L4 to 50-master.conf and only keep the log_level in 50-api.conf

That kinda defeats the purpose of having a config file for the master, and different one for the api. If we have common options, they should either be duplicated, or in a file that's common to both IMO

PR #173, opened 2018-04-26
Run activate.sh from admin-node-init.service during the first boot

thkukuk:
Currently we have a special YaST module to run active.sh after installation. To make things easier and to unify our configuration processes, I propose to run the script by a systemd service once at the first boot.
This solution is compatible with the current SLE12 based code as with the new Factory/SLE15 code.

### DIFF
--- a/activate.sh
+++ b/activate.sh
@@ -3,6 +3,11 @@
 # this script WILL BE RUN ONLY ONCE, after the installation
 # this script WILL NOT RUN during/after upgrades
 
+DIDRUNFILE=/var/lib/misc/caasp-admin-node-init-did-run
+
+# Don't run this script a second time
+test -f ${DIDRUNFILE}  && exit 0
+
 # Make sure that the controller node looks for the local pause image
 # TODO: remove this as soon as possible. As an idea, we could use a systemd drop-in unit.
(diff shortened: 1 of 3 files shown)


jordimassaguerpla:
@thkukuk : is this needed for CaaSP v3 (based on sle12)? We are very close to feature freeze and I was wondering if this could already be for the next CaaSP version.

thkukuk:
On Fri, Apr 27, Jordi Massaguer Pla wrote:

> @thkukuk : is this needed for CaaSP v3 (based on sle12)? We are very close to
> feature freeze and I was wondering if this could already be for the next CaaSP
> version.

It's a hard requirement for Kubic/CaaSP v4 only, on CaaSP v3 it would
fix the bug, that activate.sh is called a second or third time if the
cloud-init configuration was modified by an admin. So could become
relevant for Public Cloud.


-- 
Thorsten Kukuk, Distinguished Engineer, Senior Architect SLES & CaaSP
SUSE LINUX GmbH, Maxfeldstr. 5, 90409 Nuernberg, Germany
GF: Felix Imendoerffer, Jane Smithard, Graham Norton, HRB 21284 (AG Nuernberg)

PR #174, opened 2018-05-02
WIP: respect kubic when configuring the pause image

MaximilianMeister:
feature#kubic-integration

### DIFF
--- a/activate.sh
+++ b/activate.sh
@@ -5,8 +5,15 @@
 
 # Make sure that the controller node looks for the local pause image
 # TODO: remove this as soon as possible. As an idea, we could use a systemd drop-in unit.
+
+if hostnamectl | grep -q [Kk]ubic; then
+  DIST=kubic
+else
+  DIST=sles12
+fi
+
 if ! grep "pod-infra-container-image" /etc/kubernetes/kubelet &> /dev/null; then
-  sed -i 's|^KUBELET_ARGS="|KUBELET_ARGS="--pod-infra-container-image=sles12/pause:1.0.0 |' /etc/kubernetes/kubelet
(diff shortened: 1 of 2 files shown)


activate.sh
```diff
@@ -5,8 +5,15 @@
 
 # Make sure that the controller node looks for the local pause image
 # TODO: remove this as soon as possible. As an idea, we could use a systemd drop-in unit.
+
+if hostnamectl | grep -q [Kk]ubic; then
```
kiall: This seems brittle, what if I run CaaSP with a hostname of fookubicfoo - this will match. We'd be better off pulling dist info out of `lsb-release -a`.

admin-node-setup.sh
```diff
@@ -39,6 +39,12 @@ for i in $(ls $images_dir/sles*.tag $images_dir/kubic*.tag);do
 
     echo "$0: Setting $image_name:$tag into manifests"
     sed -i -e "s%$image_name:__TAG__%$image_name:$tag%g" $tmp_dir/*.yaml
+
+    if [[ $image_name =~ pause ]]; then
+      kubelet_config=/etc/kubernetes/kubelet
+      echo "$0: Setting $image_name:$tag into kubelet config"
+      sed -i -e "s%$image_name:__TAG__%$image_name:$tag%g" $kubelet_config
```
kiall: This will only replace the tag once, which is no worse than before, but still will only happen once. Something like this maybe?

    sed -i -e "s| --pod-infra-container-image=.* | --pod-infra-container-image=$DIST/pause:__TAG__ |" /etc/kubernetes/kubelet

This would also allow removing any of the pause stuff from activate.sh - best to keep all in 1 place.

MaximilianMeister:
i wasnt aware this is already done at build time on `caasp-container-manifests` - it's likely a devsetup only issue with our injected manifests ... closing

PR #175, opened 2018-05-03
[WIP] drop MySQLdb specific env

MaximilianMeister:
this only affects kubic for now where we use PyMySQL instead of
MySQLdb

salt#mysql-unix-socket

### depends on
switch to salt `2018.3` and therefore switch to `PyMySQL`


ereslibre:

PR #176, opened 2018-05-03
Run Velum's `bin/init` as an init container

ereslibre:
This will ensure that when the dashboard, api and event processor start,
the database has been created and/or migrated at all times. This avoids
the weird situation in which the api and event-processor can start while
the dashboard was still migrating the database, causing side-effects if
Rails already cached some model attributes on other processes.

Fixes: bsc#1091843

Depends on: https://github.com/kubic-project/velum/pull/506, so the `bin/init` script performs initialization logic and exits, instead of running the puma server.
Depends on: https://github.com/kubic-project/automation/pull/368

### DIFF
--- a/manifests/public.yaml
+++ b/manifests/public.yaml
@@ -115,6 +115,38 @@ spec:
     volumeMounts:
     - mountPath: /infra-secrets
       name: infra-secrets
+  - name: velum-dashboard-init
+    image: sles12/velum:__TAG__
+    env:
+    - name: RAILS_ENV
+      value: production
+    - name: VELUM_SECRETS_DIR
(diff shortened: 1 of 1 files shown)


ereslibre:
I have to validate that this approach works as is.

ereslibre:
[Bug 1091843 - When updating from 2.0 the `update all nodes` link disappears](https://trello.com/c/DnZfae7N/842-bug-1091843-when-updating-from-20-the-update-all-nodes-link-disappears)

ereslibre:
It's working.

ereslibre:
We need superpowers to merge this one. @flavio, @kiall?

PR #179, opened 2018-05-08
[3.0] Run Velum's `bin/init` as an init container

ereslibre:
This will ensure that when the dashboard, api and event processor start,
the database has been created and/or migrated at all times. This avoids
the weird situation in which the api and event-processor can start while
the dashboard was still migrating the database, causing side-effects if
Rails already cached some model attributes on other processes.

Fixes: bsc#1091843
(cherry picked from commit 2dec06e40227abdb847e376ca0dbaf5ddfc10f69)

### DIFF
--- a/manifests/public.yaml
+++ b/manifests/public.yaml
@@ -115,6 +115,38 @@ spec:
     volumeMounts:
     - mountPath: /infra-secrets
       name: infra-secrets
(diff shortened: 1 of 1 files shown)


ereslibre:
We need superpowers to merge this one. @flavio, @kiall?

PR #183, opened 2018-05-24
mount velum branding

MaximilianMeister:
feature#branding

~~still depends on the availability of the [velum-branding package](https://github.com/kubic-project/velum-branding) in the product~~ should be in the devel image now

## Needs to be merged with

https://github.com/kubic-project/velum/pull/550


MaximilianMeister:
just tested it in combination with https://github.com/kubic-project/velum/pull/550 and it looks good now that we have the branding in the devel images installed by default

PR #184, opened 2018-05-28
fix image name for CaaSP4

jordimassaguerpla:
We have changed the naming of the images. Instead of calling them
sleXY/image_name, we call them caasp/image_name, so we don't have to
create a new package each time we change the base image.

### DIFF
--- a/packaging/suse/make_spec.sh
+++ b/packaging/suse/make_spec.sh
@@ -42,7 +42,7 @@ cat <<EOF > ${NAME}.spec
 %endif
 
 %if 0%{?suse_version} == 1500 && !0%{?is_opensuse}
-  %define _base_image sles15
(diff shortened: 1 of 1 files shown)


jordimassaguerpla:
Sorry, caasp4 does not require any rpm

jordimassaguerpla:
Note that we need to create another fix for changing the image name so it looks for the image in the right registry.

packaging/suse/make_spec.sh
```diff
@@ -64,22 +64,6 @@ Source:         ${SAFE_BRANCH}.tar.gz
 Requires:       container-feeder
```
davidcassany: Shouldn't it be something like

```
%if 0%{?sle_version} < 150000
...
%endif
```

PR #188, opened 2018-05-30
mount static error pages from velum-branding (bsc#1080636)

MaximilianMeister:

### depends on 
https://github.com/kubic-project/velum-branding/pull/7

### DIFF
--- a/config/haproxy/haproxy.cfg
+++ b/config/haproxy/haproxy.cfg
@@ -22,6 +22,7 @@ listen velum
         default-server inter 10s fall 3
         balance roundrobin
         server velum unix@/var/run/puma/dashboard.sock
+        errorfile 503 /etc/caasp/haproxy/errors/503.html.http
 
 listen velum-api
         bind 127.0.0.1:444 ssl crt /etc/pki/private/velum-bundle.pem ca-file /etc/pki/ca.crt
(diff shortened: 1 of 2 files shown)


MaximilianMeister:
even though it's a bug we can't backport it as we dont have the `velum-branding` package in `v3`

flavio:
Should we force merge this PR or is there something missing from the image? I don't think the failure on Jenkins is related with this PR...

MaximilianMeister:
> Should we force merge this PR or is there something missing from the image? I don't think the failure on Jenkins is related with this PR...

let's wait for the branding issue to be fixed and then continue here. i'll set it to WIP

MaximilianMeister:
I;ve just tested it, and it shows the nice page now

config/haproxy/haproxy.cfg
```diff
@@ -11,6 +11,8 @@ defaults
         timeout client 120000
         timeout server 120000
 
+errorfile 503 /etc/caasp/haproxy/errors/503_database_not_ready.html.http
```
ereslibre: Maybe we can use a rename of the file... because it's not that the database is not ready, it might be a myriad of different reasons.

Aside from that, could we move this inside the `defaults` block?

MaximilianMeister: renamed and added the errorfile in the `listen velum` block

PR #189, opened 2018-05-31
Configure `innodb_log_file_size` to a 128M limit

ereslibre:
When a salt output is big enough, mysql will refuse to insert the offending
row for being too big, with an error:

```
[ERROR   ] Could not store events - returner 'mysql.event_return' raised exception:
(1118, 'The size of BLOB/TEXT data inserted in one transaction is greater than 10% of redo log size.
Increase the redo log size using innodb_log_file_size.')
```

Whatever we set as `innodb_log_file_size` will be an arbitrary number that will
eventually be flooded if the cluster is big enough, or if salt is noisy enough.
Given a cluster size, this can suddenly fail if we add more states (thus, increasing
salt's output). Obviously, given the same salt states, we can also reach this limit
by increasing the cluster size.

There is not a definitive fix for this issue, all we can do for now (without a proper
refactor of the way we integrate salt and velum) is to ensure that with todays salt
states we can reach a certain number of nodes. As said, this can no longer be true if
we add more salt states and we reach again the limit for the same cluster size.

Fixes: bsc#1095335

Depends on: https://github.com/kubic-project/container-images/pull/31

### DIFF
--- a/manifests/private.yaml
+++ b/manifests/private.yaml
@@ -38,17 +38,13 @@ spec:
       value: "true"
     - name: MYSQL_ROOT_PASSWORD_FILE
       value: /var/lib/misc/infra-secrets/mariadb-root-password
+    - name: MYSQLD_CONFIG
+      value: "skip-networking;max_allowed_packet=16M;innodb_log_file_size=128M"
     volumeMounts:
     - mountPath: /var/lib/mysql
       name: mariadb-data
     - mountPath: /var/run/mysql
       name: mariadb-unix-socket
-    - mountPath: /etc/my.cnf.d/skip-networking.cnf
-      name: mariadb-config-skip-networking
-      readOnly: True
-    - mountPath: /etc/my.cnf.d/max-allowed-packet.cnf
-      name: mariadb-config-max-allowed-packet
-      readOnly: True
     - mountPath: /var/lib/misc/infra-secrets
       name: infra-secrets
       readOnly: True
@@ -59,12 +55,6 @@ spec:
   - name: mariadb-unix-socket
     hostPath:
       path: /var/run/mysql
-  - name: mariadb-config-skip-networking
-    hostPath:
-      path: /usr/share/caasp-container-manifests/config/mysql/skip-networking.cnf
-  - name: mariadb-config-max-allowed-packet
-    hostPath:
-      path: /usr/share/caasp-container-manifests/config/mysql/max-allowed-packet.cnf
   - name: infra-secrets
     hostPath:
       path: /var/lib/misc/infra-secrets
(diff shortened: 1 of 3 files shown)


ereslibre:
[Bug 1095335 - Bootstrapping hangs forever](https://trello.com/c/3NEYOMJm/991-bug-1095335-bootstrapping-hangs-forever)

ereslibre:
> Can we at least make the size of the log file configurable via an environment variable? By doing that some users could quickly change that to fit their needs.

Hmmm, I am not sure we can really do this. The config file will live in a RO filesystem and we don't really suggest users to remount the filesystem as RW. We can set that as an envvar, but I'm not sure if it's really worth it. What do you think?

kiall:
> Can we at least make the size of the log file configurable via an environment variable? By doing that some users could quickly change that to fit their needs.

Right - I'm not sure I see how an environment variable would make this dynamic? The env var would be set in the manifest, which we're going to override in salt time+time again.

manifests/private.yaml
```diff
@@ -49,6 +49,9 @@ spec:
     - mountPath: /etc/my.cnf.d/max-allowed-packet.cnf
       name: mariadb-config-max-allowed-packet
       readOnly: True
+    - mountPath: /etc/my.cnf.d/innodb-log-file-size.cnf
+      name: mariadb-config-innodb-log-file-size
+      readOnly: True
```
kiall: I would *love* to see us stop mounting stuff into the containers ;) Given this is static, can we just bake it into the image?

And, if we do add a env var - I'd still prefer to see it in the image, with the entrypoint rendering the file based on the env var / default value.

ereslibre: I agree this is not the best way, and that we should stop doing this, but I find more concerning having some settings in a file and some others being rendered from an envvar, and this is why I went with this option. It's as static as everything under https://github.com/kubic-project/caasp-container-manifests/tree/master/config/mysql.

Summing up: I have no objections in adding this to the image, but I think if we do that we should also move the settings under https://github.com/kubic-project/caasp-container-manifests/tree/master/config/mysql to basically do the same thing; so at least we are consistent.

kiall: Makes sense. LGTM then. We really need to clean it all up at some point though :)

ereslibre: The change is a bit messy, this now depends on https://github.com/kubic-project/container-images/pull/31 for 4.0, and the backport needs a SR created against IBS, so I'll wait until https://github.com/kubic-project/container-images/pull/31 gets reviewed to continue with the SR.

pgonin:
The change unblocked the scaling tests and more nodes could be added to cluster (beyond 140 nodes)

PR #192, opened 2018-06-07
Do not mount the public directory or the compiled assets get hidden in

jordimassaguerpla:
the iso

Mounting the public directory from the host, was "hidding" the assets
files present in the public directory and so our images were broken.

This was only happening in the iso and not in our dev environments
because in production we have the assets precompiled and we don't
compile them on the fly, since we don't have a js engine installed.

### DIFF
--- a/manifests/public.yaml
+++ b/manifests/public.yaml
@@ -589,9 +589,6 @@ spec:
   - name: velum-dist-name
     hostPath:
       path: /usr/share/velum/PRODUCT
-  - name: velum-static-pages
-    hostPath:
-      path: /usr/share/velum/public
   - name: velum-images
     hostPath:
       path: /usr/share/velum/images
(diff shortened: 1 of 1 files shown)


jordimassaguerpla:
This goes with https://github.com/kubic-project/velum-branding/pull/5

jordimassaguerpla:
Also with https://github.com/kubic-project/velum/pull/579

manifests/public.yaml
```diff
@@ -589,9 +589,6 @@ spec:
   - name: velum-dist-name
     hostPath:
       path: /usr/share/velum/PRODUCT
-  - name: velum-static-pages
-    hostPath:
-      path: /usr/share/velum/public
```
MaximilianMeister: we could keep this e.g. with `path: /usr/share/velum/public/static-pages` and then set `paths["public"] = "static-pages"` in https://github.com/kubic-project/velum/blob/master/config/application.rb

manifests/public.yaml
```diff
@@ -315,7 +315,7 @@ spec:
     - mountPath: /srv/velum/public
```
MaximilianMeister: this mount would need to be dropped too

PR #197, opened 2018-07-04
Use random CA serial numbers

inercia:
We should generate a random CA serial number. According to the CA/Browser Forum Baseline Requirements section 7.1: _"CAs SHOULD generate non‐sequential Certificate serial numbers that exhibit
at least 20 bits of entropy."_. In general it is considered a good practice to use a random number instead of a constant...

feature#security

### DIFF
--- a/gen-certs.sh
+++ b/gen-certs.sh
@@ -9,10 +9,15 @@ CITY=${CITY:-Nuremberg}
 STATE=${STATE:-Bavaria}
 COUNTRY=${COUNTRY:-DE}
 
-DIR="/etc/pki"
+DIR="${DIR:-/etc/pki}"
 CERTS="$DIR/_certs"
 PRIVATEDIR="$DIR/private"
 WORK="$DIR/_work"
+
+
+random_serial() {
+    xxd -l 16 -p /dev/random
+}
 
 genca() {
     [ -f $PRIVATEDIR/ca.key ] && [ -f $DIR/ca.crt ] && return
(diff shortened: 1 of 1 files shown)


gen-certs.sh
```diff
@@ -9,11 +9,16 @@ CITY=${CITY:-Nuremberg}
...
 
-DIR="/etc/pki"
+DIR="${DIR:-/etc/pki}"
 CERTS="$DIR/_certs"
 PRIVATEDIR="$DIR/private"
 WORK="$DIR/_work"
 
+
+random_serial() {
+    python -c 'import random ; print(random.randint(1, 2**159))'
```
kiall: What is the max acceptable serial number? If it's 2**159 - then, if we choose that (or anything close to it..) we'll be in trouble. We'll need to ensure we have enough room for ever cert + renewal we'll ever issue!

kiall: We can easily avoid python for this.. something like `$(((RANDOM%10000000)+1))` - with a suitably high number in place of the `10000000`.

inercia: It seems that `"The range of returned values is 0 to 32767"` for `$RANDOM`. Taking into account that the serial can be up to `2^160`, maybe it is not a big deal but it seems slightly short... 

However, I've found [this](https://unix.stackexchange.com/questions/157250/how-to-efficiently-generate-large-uniformly-distributed-random-integers-in-bas) with some other ways to generate random numbers. I'll find some other way.

PR #198, opened 2018-07-23
re-introduce config flag

MaximilianMeister:
with the kubelet-config.yaml we need it again

kubelet#config

this should fix the devsetup again

### DIFF
--- a/admin-node-setup.sh
+++ b/admin-node-setup.sh
@@ -9,9 +9,6 @@ fi
 # Turn off current swaps if any
 /usr/sbin/swapoff -a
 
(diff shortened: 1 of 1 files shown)


MaximilianMeister:
NOTE: this is not yet enough, as kubelet still sets some unwanted defaults:

```
F0723 06:47:59.750236    2788 server.go:233] failed to run Kubelet: no client provided, cannot use webhook authorization
```

MaximilianMeister:
@mjura will send a k8s package fix, then we can re-run this

MaximilianMeister:
https://build.opensuse.org/request/show/624656

PR #199, opened 2018-08-23
Fix bsc#1099045 adds annotation to use docker/default seccomp profile

chentex:
This annotation will make haproxy use the docker/default seccomp profile.

Signed-off-by: Vicente Zepeda Mas <vzepedamas@suse.com>


chentex:
[Bug 1099045](https://trello.com/c/Q6MCDKkO/1138-bug-1099045)

PR #200, opened 2018-09-19
WIP add images from registry

jordimassaguerpla:
We are not using rpms anymore to deploy containers for caasp4(SLE15).
However, we would like to choose which images we use, so we will use a
patch instead of a sed command.

### DIFF
--- a/packaging/suse/set_caasp_images.patch
+++ b/packaging/suse/set_caasp_images.patch
@@ -1,0 +1,178 @@
+diff --git a/activate.sh b/activate.sh
+index ccd9d84..c002ece 100755
+--- a/activate.sh
++++ b/activate.sh
+@@ -11,7 +11,7 @@ test -f ${DIDRUNFILE}  && exit 0
+ # Make sure that the controller node looks for the local pause image
+ # TODO: remove this as soon as possible. As an idea, we could use a systemd drop-in unit.
+ if ! grep "pod-infra-container-image" /etc/kubernetes/kubelet &> /dev/null; then
+-  sed -i 's|^KUBELET_ARGS="|KUBELET_ARGS="--pod-infra-container-image=sles12/pause:1.0.0 |' /etc/kubernetes/kubelet
++  sed -i 's|^KUBELET_ARGS="|KUBELET_ARGS="--pod-infra-container-image=registry.suse.de/devel/casp/head/controllernode/images_container_base/sles12/pause:1.0.0 |' /etc/kubernetes/kubelet
+ fi
+ 
+ # Make sure etcd listens on 0.0.0.0
+diff --git a/manifests/haproxy.yaml b/manifests/haproxy.yaml
+index d10f01c..9718ad5 100644
+--- a/manifests/haproxy.yaml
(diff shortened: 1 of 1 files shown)


jordimassaguerpla:
I don't like very much this thing but I don't know how to make it work with kubic and caasp. I wanted to keep compatibility with sle12 since we don't yet have the sle15 image, but this is a "nice to have" thing.

@davidcassany @bergmannf : any ideas on how to do this differently?

jordimassaguerpla:
what if we had yaml files for each version? like public.yaml.caasp4, private.yaml.caasp4, public.yaml.caasp3, private.yaml.caasp3, public.yaml.kubic, etc...

And then in the package we install one or the other depending on the build architecture. How does it sound?

bergmannf:
I think using the concrete versions might be a problem - I don't know how often those change, but whenever they do we have to change the `patch`.

Maybe there is a way to use one of those `source-services` to generate the patched-in images to use the latest tag available under a certain image-name from registry.suse.de?

Also - if we do it like this, I think using the same approach in `kubernetes-salt` makes sense, to use a placeholder in case of using registry.suse.de and patching the concrete images in during build time.

jordimassaguerpla:
Having some automation in place would be great. Actually, once we were discussing with @davidcassany about getting the information from the build service by generating some metadata packages with each image build, and requiring this metadata when building c-c-m package (and k8s-salt).

The key point is to know if the tags will change that much.

Thus, will velum image be tagged on velum version, which changes on each commit, or will the velum image be tagged with a more stable number, as 0.0?

jordimassaguerpla:
https://github.com/kubic-project/rfc/pull/7

jordimassaguerpla:
I agree using the patch is not a good approach.

If we were using some kind of obs trick to get the tag, should we leave some defaults in case we don't use the RPM?

@ereslibre @MaximilianMeister : Are the development environments using the manifests files without the RPM?

jordimassaguerpla:
@bergmannf : Right now we are using the images based on sle12, like registry.suse.de/devel/casp/head/controllernode/images_container_base/sles12/velum:0.0, which have a tag that does not change.

What if we used this as a default until we have the automation in place?

jordimassaguerpla:
If we were using, as a first step, the ones based in sle12: registry.suse.de/devel/casp/head/controllernode/images_container_base/sles12/velum:0.0 we could easily do that by setting the _base_image variable to registry.suse.de/devel/casp/head/controllernode/images_container_base/sles12 in the spec file.

Then on a second step, we could use the obs mechanism.

How does it sound?

bergmannf:
@jordimassaguerpla  I think setting the `_base_image` would be cleaner than the patch - I didn't know that we were already using `sed` in the spec-file. To get something working I would be fine with that.

AFAIK the dev environment is using the packaged `rpm` files right now (except for `velum` which is modified on the fly - but I think that is also adjusted on the fly, so we might have to change the `automation` around that to correspond to the new image names): but let's wait for @MaximilianMeister or @ereslibre to chime in on that part.

PR #201, opened 2018-09-20
use the sles12 images from registry.suse.de

jordimassaguerpla:
This is a first step towards using the registry.
We will use an internal one in registry.suse.de
Later, we will use the registry.suse.com for caasp and
registry.opensuse.org for kubic, however, we have to figure out the
tagging strategy if we want to use kubic images or images based on
sle15.

Thus, for now, we start using the sles12 images which have a fixed tag.

### DIFF
--- a/packaging/suse/make_spec.sh
+++ b/packaging/suse/make_spec.sh
@@ -42,7 +42,8 @@ cat <<EOF > ${NAME}.spec
 %endif
 
 %if 0%{?suse_version} == 1500 && !0%{?is_opensuse}
-  %define _base_image caasp
+  # Use the sles12 images from the registry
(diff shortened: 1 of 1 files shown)


bergmannf:
@davidcassany Ah good clarification, I didn't realize the `sle` vs `suse` prefix - in which case I would assume using `sle_version` everywhere, where we want to detect sle15 would be the right way to go.

jordimassaguerpla:
Using sle_version may not be very reliable https://lists.opensuse.org/opensuse-buildservice/2014-11/msg00026.html 

I am still confused when it comes to detect the suse/sles version, to be honest.

PR #202, opened 2018-10-03
Use 3.0 images instead of head images

jordimassaguerpla:
Because the images from head will be removed to avoid duplications. The ones
we are using is the ones from 3.0 project, cause those ones cannot be
removed.


jordimassaguerpla:
@bergmannf : true. I always forget about that. Let me create another PR.
@davidcassany : This is for sle12sp3 images. How should this be done for sle15?

PR #203, opened 2018-10-15
Store the registry configuration as part of the manifests.

bergmannf:
This information will be modified and set during the build-time of the packages,
depending on the distribution it will be released to:

on SLE15 the information will contain the registry and set the use_registry to
true, on SLE12 the use_registry will remain false and use the images from
container-feeder.

### DIFF
--- a/packaging/suse/0001-patch-in-version-numbers-for-images.patch
+++ b/packaging/suse/0001-patch-in-version-numbers-for-images.patch
@@ -1,0 +1,179 @@
+From d9a6768a28becc3fc951bbd3c06ff01d8632535c Mon Sep 17 00:00:00 2001
+From: Florian Bergmann <fbergmann@suse.de>
+Date: Fri, 12 Oct 2018 17:27:28 +0200
+Subject: [PATCH] patch in version numbers for images
+
+---
+ manifests/haproxy.yaml |  2 +-
+ manifests/private.yaml |  4 ++--
+ manifests/public.yaml  | 28 ++++++++++++++--------------
+ 3 files changed, 17 insertions(+), 17 deletions(-)
+
+diff --git a/manifests/haproxy.yaml b/manifests/haproxy.yaml
+index ada52aa..24f4144 100644
+--- a/manifests/haproxy.yaml
++++ b/manifests/haproxy.yaml
+@@ -20,7 +20,7 @@ spec:
+       operator: "Exists"
+   containers:
+     - name: haproxy
+-      image: sles12/haproxy:__TAG__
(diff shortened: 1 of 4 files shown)


packaging/suse/make_spec.sh
```diff
@@ -101,13 +101,16 @@ for file in manifests/*.yaml; do
   # fix image name
   sed -e "s|image:[ ]*sles12/\(.*\):|image: %{_base_image}/\1:|g" -i %{buildroot}/%{_datadir}/%{name}/\$file
 done
+# Install registry-configuration file
+install -d %{buildroot}/%{_datadir}/%{name}/config/registry
+install -D -m 0644 config/registry/registry-config.yaml %{buildroot}/%{_datadir}/%{name}/config/registry/registry-config.yaml
```
MaximilianMeister: we should probably add the new folder/file as well in the `%files` section

bergmannf: I think the file is covered by `%{_datadir}/%{name}/*` already.

packaging/suse/0001-patch-in-version-numbers-for-images.patch
```diff
@@ -0,0 +1,179 @@
...
++    image: sles12/velum:0.0
+     command: ["sh", "-c", "umask 377;
+                            if [ ! -f /infra-secrets/saltapi-password ]; then
+                              head -n 10 /dev/random | base64 | head -n 10 | tail -n 1 > /infra-secrets/saltapi-password;
+@@ -55,7 +55,7 @@ spec:
+     - mountPath: /infra-secrets
+       name: infra-secrets
+   - name: salt-minion-key-generation
+-    image: sles12/salt-master:__TAG__
++    image: sles12/salt-master:2018.3.0
```
davidcassany: Shouldn't we use `2018.3` only to be consistent? A very minor issue, I am just wondering if we should or not be strict on that.

jordimassaguerpla: I think is a good point but this images are already released :) . We could fix that for the next images, but it is too late for the sles12 ones.

bergmannf: I think the salt-images were always the odd-ones - the `2016` ones are already using this kind of format: `salt-master:2016.11.4`.

It also reflects the versioning our salt-team uses for our internal salt-releases (and I think a change from `2018.3.0` to let's say `2018.3.2` seems to be a *bigger* change already), so maybe keeping it might be worthwhile.

PR #206, opened 2018-11-15
[BSC#1111880] Change internal LDAP group name to prevent conflicts

nanoscopic:
Change internal LDAP group name to prevent external auth conflicts

The internal LDAP group name for the Velum admin user was 'Administrators'. Once external auth is enabld to a companies LDAP server, this causes anyone in an Administrators group there also to gain permission to be a cluster admin.

Changing the name of the group to CaaSP-Administrators to prevent this conflict from occurring and it to be explicity obvious what the group is intended to administrate.

### DIFF
--- a/manifests/public.yaml
+++ b/manifests/public.yaml
@@ -289,7 +289,7 @@ spec:
     - name: LDAP_BASE_DN
       value: dc=infra,dc=caasp,dc=local
     - name: LDAP_ADMIN_GROUP_DN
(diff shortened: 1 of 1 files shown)


dannysauer:
Based on a quick grep, the `Administrators` string appears hard-coded in http://github.com/kubic-project/velum in a few places.  One appears to just be a fall-back if not specified, but it'd be a good idea to double-check on that.

Otherwise, this LGTM - once you (or someone) has figured out how this gets applied to an existing cluster in an update situation and noted that either here or in the bugzilla record. :)

PR #207, opened 2018-11-28
use /var/run/salt/master on host as socket dir

jgleissner:
Use /var/run/salt/master instead of /var/run/salt/master/sock-dir, as salt-cloud on the admin host expects the salt master socket files in that directory to send cloud events.

This is required to make https://github.com/kubic-project/velum/pull/700 work properly.

### DIFF
--- a/manifests/public.yaml
+++ b/manifests/public.yaml
@@ -541,7 +541,7 @@ spec:
       path: /usr/share/salt/kubernetes
   - name: salt-sock-dir
     hostPath:
-      path: /var/run/salt/master/sock-dir
+      path: /var/run/salt/master
   - name: salt-minion-ca-cache
     hostPath:
       path: /var/lib/misc/salt/cache/salt-minion-ca


manifests/public.yaml
```diff
@@ -541,7 +541,7 @@ spec:
       path: /usr/share/salt/kubernetes
   - name: salt-sock-dir
     hostPath:
-      path: /var/run/salt/master/sock-dir
+      path: /var/run/salt/master
```
ereslibre: Why is this change required? In a 3.0 cluster I see the following on the admin node:

```
admin:~ # ls -la /var/run/salt/master/sock-dir
total 0
drwxr-xr-x 2 root root 120 Nov 29 11:37 .
drwxr-x--- 3 salt salt  60 Nov 29 11:37 ..
srw-rw-rw- 1 root root   0 Nov 29 11:37 master_event_pub.ipc
srw------- 1 root root   0 Nov 29 11:37 master_event_pull.ipc
srw------- 1 root root   0 Nov 29 11:37 publish_pull.ipc
srwxr-xr-x 1 root root   0 Nov 29 11:37 workers.ipc
```
Is this different in the cloud case?

jgleissner: Actually, we don't really need this, as the path can be set by creating /etc/salt/master on the host containing `sock_dir=/var/run/salt/master/sock-dir`. There's no actual master on the admin host itself, but it'll change the directory salt-cloud looks for the master ipc sockets.

I was under the impression that the path was constructed somehow from the path that was configured in the minion, judging from a few `os.path.join(__opts__['sock_dir'], 'master'))` in `salt/utils/cloud.py` (and those particular events are probably actually broken, but fortunately we're not interested in them). The events triggered by the cloud driver modules use the proper sock_dir so they should be fine. I'll add the necessary master config file to the cloud admin image builds, and this PR can be closed. Sorry about the noise.

jgleissner: Actually, it's probably better if admin-node-setup.sh sets configures the path. That way it'll also work for upgrades, whereas simply including it in the image won't. I'll open a PR.

PR #211, opened 2019-01-17
point to registry.suse.com for the container images

MaximilianMeister:

@jordimassaguerpla another thing that i discovered, which looks like a bug is https://github.com/kubic-project/caasp-container-manifests/blob/master/packaging/suse/make_spec.sh#L68-L82

even before the change of this PR, and expanded `%{_base_image}` would have looked like this in the requires:

```rpm
Requires:       registry.suse.de/devel/casp/3.0/controllernode/images_container_base/sles12-pause-image >= 2.0.0
```

is this correct?

### DIFF
--- a/packaging/suse/make_spec.sh
+++ b/packaging/suse/make_spec.sh
@@ -43,7 +43,7 @@ cat <<EOF > ${NAME}.spec
 
 %if 0%{?suse_version} == 1500 && !0%{?is_opensuse}
   # Use the sles12 images from the registry
-  %define _base_image registry.suse.de/devel/casp/3.0/controllernode/images_container_base/sles12
+  %define _base_image registry.suse.com/sles12
 %endif
 
 %if 0%{?is_opensuse} && 0%{?suse_version} > 1500


jordimassaguerpla:
@MaximilianMeister On your question about caasp-container-manifests. See that this code does not apply for SLE15 (caasp4) https://github.com/kubic-project/caasp-container-manifests/blob/master/packaging/suse/make_spec.sh#L64

This will only apply for caaspv3 or kubic, which the base_image is fine. So I don't see any error there.

jordimassaguerpla:
@MaximilianMeister However, could you do the same change you are doing here in https://github.com/kubic-project/caasp-container-manifests/blob/master/packaging/suse/make_spec.sh#L46? So that we use the images from registry.suse.com

MaximilianMeister:
> @MaximilianMeister However, could you do the same change you are doing here in https://github.com/kubic-project/caasp-container-manifests/blob/master/packaging/suse/make_spec.sh#L46? So that we use the images from registry.suse.com

should be addressed already here. or did you mean https://github.com/kubic-project/salt/blob/master/packaging/suse/make_spec.sh#L47 ? I already have a commit for that, but i'll send it together with some more salt changes related to the v4 migration

PR #221, opened 2019-02-20
Use caasp v4 images

jordimassaguerpla:


jordimassaguerpla:
This needs to go together with https://github.com/kubic-project/salt/pull/747

jordimassaguerpla:
@MaximilianMeister : will it be ok for you that we merge both Requests? I don't think  tests can pass separately...

MalloZup:
https://github.com/SUSE/caasp-jenkins-library/pull/276

MaximilianMeister:
> will it be ok for you that we merge both Requests? I don't think tests can pass separately...

yes, if you can force merge go ahead. image tags are all correct

MaximilianMeister:
@jordimassaguerpla can we now do it without a patch? see https://github.com/kubic-project/caasp-container-manifests/pull/222

MaximilianMeister:
now handeled in https://github.com/kubic-project/caasp-container-manifests/pull/222

PR #224, opened 2019-03-01
Add missing kube-apiserver-haproxy-bundle.pem

dannysauer:
I forgot to add this file mount to _this_ manifest, so it gets changed on reboot and breaks haproxy after salt changes the haproxy config during orchestration.  Merging this should fix the CI failure on kubic-project/salt#721

Probably should open an issue to have Salt maintain the /usr/share/caasp file when it updates the "real" file.  Or an issue to get rid of management of the same files by two processes...


stefsuse:
Change seems reasonable to me but I'll let someone else hit approve. 

Agreed that two processes should probably not manage the state of one file.. want to file the issue for that?

nanoscopic (approved):
Looks good.

PR #222, opened 2019-02-26
adapt container image paths to be from the registry by default

MaximilianMeister:
Signed-off-by: Maximilian Meister <mmeister@suse.de>

in favor of #221 which uses a patch

PR #226, opened 2019-03-05
adapt container image paths to be from the registry by default

MaximilianMeister:
instead of https://github.com/kubic-project/caasp-container-manifests/pull/222 to make it easier to test with other dependent PR's


MaximilianMeister:
if https://github.com/kubic-project/caasp-container-manifests/pull/228 takes precedence, i will have to rebase this PR after merging it

the reason is that we can only test this change here when the manifests in the qcow image itself are not already patched

jordimassaguerpla (changes requested):
You can get rid of _base_image in the spec file. Other than than, lgtm

MaximilianMeister:
> You can get rid of _base_image in the spec file. Other than than, lgtm

fixed, thanks!

jordimassaguerpla (approved):

PR #228, opened 2019-03-05
get rid of the patch

MaximilianMeister:
it creates problems with testing a follow up PR where we
switch to the registry images dynamically (via __TAG__
expansion)

- drop 0001-patch-in-version-numbers-for-images.patch


jordimassaguerpla:
> Can we have the tags in the manifests? Otherwise this will break the images.

I mean here https://github.com/kubic-project/caasp-container-manifests/tree/master/manifests and whatever the patch was changing the tags.

MaximilianMeister:
> I mean here https://github.com/kubic-project/caasp-container-manifests/tree/master/manifests and whatever the patch was changing the tags.

this is now part of https://github.com/kubic-project/caasp-container-manifests/pull/226

